PR #1, opened 2013-10-02
Update hello.py

bparrish206:


PythonCHB:
I suspectyou didn't mean to commit this particualr change.

git commit -a

can be tricky!

Note, if you have changed a file you didn't mean to, or don't want to keep, you can do:

git checkout hello.py

and you'll get the version from the repo back.

PR #2, opened 2014-10-05
Added SSCwafel directory and a test file

saschwafel:


PythonCHB:
did you mean to put in an empty file?

You may have noticed that git doesn't let you add an empty directory, but does let you add empty files.

saschwafel:
Yeah, I was trying to get a hang of this by simply adding the SSchwafel directory. However, as you might expect it wouldn't track that directory's presence as a change so I added (then removed, then re-added) the test.py file in order to commit the changes. I just wanted proof of concept that I could get this to sync.

PythonCHB:
Sounds good -- I'll merge it.

PR #4, opened 2014-10-08
Week 1 Homework for Salim

salimhamed:


Students/salim/draw_box.py
```diff
@@ -0,0 +1,74 @@
+def print_grid(size, boxes):
+
+    # adjust wrong inputs
+    size, boxes = adjust_inputs(size, boxes)
```
PythonCHB, line 4: nice use of separate functions to keep it clean.

Students/salim/draw_box.py
```diff
@@ -0,0 +1,74 @@
...
+
+    # rows of grid
+    row_count = 0
+    for i in range(size):
+
+        # if row is a solid line
+        if row_count % (size / boxes) == 0:
```
PythonCHB, line 11: you can do the "simpler":
if not row_count % (size / boxes):

(zero evaluates at False)

PR #7, opened 2014-10-08
My Python Homeworks

jerrytozhang:


jerrytozhang:
My session 1 homework

Students/Hui Zhang/session02/test-ack.py
```diff
@@ -0,0 +1,48 @@
...
+if ack(0, 4) == 5:
+	count1 = count1 + 1
+if ack(1, 4) == 6:
+	count1 = count1 + 1
+if ack(2, 4) == 11:
+	count1 = count1 + 1
+if ack(3, 4) == 125:
+	count1 = count1 + 1
+
+if count1 == 20:
```
PythonCHB, line 46: interesting way to keep track of tests!

I usually just put a lot of asserts in. But this is nice, as it keeps running when one fails.

we'll learn about "proper" test frameworks in a few weeks.

PythonCHB:
Looking good!

-Chris

PR #10, opened 2014-10-11
Session One Homework

gosunfish:


PythonCHB:
oops, this is in a session01 dir in the main dir -- you need to put it in your Students dir. Probably something like:

git mv session01 Students/YourName/

you may need to create the Students/YourName dir first.

-Chris

gosunfish:
I am completely lost with git. 

----- Original Message -----

gosunfish:
Last week I did the fork, and then the clone. 
git clone https://github.com/gosunfish/IntroToPython.git 

So if I only cloned my own fork, how can checkins or pushes or whatever be going to your area? 

When I do a git branch, it never shows up in my fork. I just really do not understand git. 

----- Original Message -----

PythonCHB:
A few notes:

commits an pushed aren't going to the main repo -- they won't until I accept the "pull request". But your rep is a "clone" of the main one, so if you put a new file somewhere, like /session01 -- then if I merge it with the mian repo that file will end up in the same place, which is not where we want it. We're using git in a bi tof a wierd way for this clas, but the priciples are the same.

It also looks like you've commited a bunch of ".idea" files -- PyCharm configuration, maybe? I supect you di da "git add *" that can be tempting, but it is all too easy to accidentally add stuff like that. IUf you google: ".idea files" you'll see that you aren't the only one with that issue. So:

a) don't "git add *" -- "git add *.py" will do it.

b) you can configure got to ignore .idea files. maybe this will help:

http://stackoverflow.com/questions/11124053/accidentally-committed-idea-directory-files-into-git

-Chris

PythonCHB:
print_grid is looking good!

Sorry for the git frustration -- it does take a while to wrap your head around.

-Chris

PythonCHB:
I"m going to close this one -- you can do a new one when you've got it cleaned up.

PR #11, opened 2014-10-11
Session02

gosunfish:


Students/gosunfish/print_grid.py
```diff
@@ -0,0 +1,97 @@
+# This was my first attempt.
+# This one prints the number of boxes horizontal and vertical based on the 'x' parameter,
+# and the size of each box is determined by the 'x' parameter as well.
+def grid1(x):
+
```
PythonCHB, line 5: So this is odd -- there are extra linefeeds in this file. Usually this happens when someone is on Windows -- it used cr and lf for a line break. But judging by the .DStore file, you're using a mac, which should use regular Unix line endings -- you may want to check your editor configuration for that.

PythonCHB:
You'll need to "git rm" the .DStore files, too -- we don't want those in version control.

don't forget to do:

"git status"

before you commit, and before you push -- that will let you know what's going on.

-Chris

PythonCHB:
closing -- resubmit when you get it cleaned up.

PR #13, opened 2014-10-12
Session01

gosunfish:


PythonCHB:
Closing this -- you can do another one when you get the .idea files cleaned up.

PR #15, opened 2014-10-13
Ian M. Davis Homework

imdavis:
Homework for UW Intro to Python


PythonCHB:
Thanks for putting in your lightning talk -- MPL is pretty nice!

I do encourage you to take a look at the "OO interface", rather than "from pylab import *" -- the pylab interface is nice for quick interactive use, and is comfortable for Matlab folks, but makes for messy, hard to re-use code when you get bigger projects going...

Students/imdavis/session01/buildgrid.py
```diff
@@ -0,0 +1,26 @@
+# a function which takes on a single argument defining the size
+# of a grid to draw
+
+def print_grid(gridsize):
+    # define the top and bottom portion of a single grid cell
+    corner = "+"
+    midtopbot = 4 * "-"
```
PythonCHB, line 7: nice use of defined constants and string multiplication

Students/imdavis/session01/drawgrid.py
```diff
@@ -0,0 +1,10 @@
+# draws a grid using the buildgrid function after prompting the
+# user for the size of the grid
+
+import buildgrid
+
+try:
+    size = int(raw_input("What size grid would you like to build? "))
+    buildgrid.print_grid(size)
+except ValueError:
```
PythonCHB, line 9: jumping ahead to Exceptions, I see -- nicely done.

PR #16, opened 2014-10-14
turning in sessions02 homework for alex galvin

ghost:
I am adding the two assignments, ack.py and series.py to my student folder

### DIFF
--- a/Students/alexg/session02/series.py
+++ b/Students/alexg/session02/series.py
@@ -1,0 +1,43 @@
+#!/usr/bin/python
+
+def fibonacci(x):
+    """Returns the nth value of the Fibonacci Series"""
+    if not x: 
+        return 0
+    elif x == 1: 
+        return 1
+    else: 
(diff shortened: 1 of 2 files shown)


Students/alexg/session02/series.py
```diff
@@ -0,0 +1,43 @@
...
+        return 2
+    elif x == 1:
+        return 1
+    else:
+        return lucas(x-1)+lucas(x-2)
+
+def sum_series(n,x=0,y=1):
+    """Returns the nth value of either series, or creates a new one"""
+    if x==0 and y==1:
+        return fibonacci(n)
```
PythonCHB, line 24: This makes sense, but as the rest of the code can compute the Fibonacci and Lucas series, you may as well jsut use that. i.e. the general form usually doesn't call a more specific version, rather the other way around. Of course, that all got confused by by the fact that we asked you to write the other two first...

PR #19, opened 2014-10-14
This includes last week's homework (abandoned) and THIS week's homework (complete)

saschwafel:
Includes Ackermann.py, Series.py, and misc. class files (including notes from both class sessions)


Students/SSchwafel/session02/ack.py
```diff
@@ -0,0 +1,78 @@
...
+
+print """
+
+This program performs Ackermann's Function, as defined by: 
+
+A(m,n) =
+
+    n+1 if m = 0
+
+    A(m-1) if m > 0 and n = 0
```
PythonCHB, line 13: nice use of conditional expressions.

Students/SSchwafel/session02/series.py
```diff
@@ -0,0 +1,83 @@
...
+    elif n > 1:
+        return lucas(n-1) + lucas(n-2)
+
+#print lucas(user_number)
+#print fibonacci(user_number)
+
+def sum_series(n, optional1 = 0, optional2 = 1):
+    """This function takes one required argument and two optional ones. 
+
+    If the optional arguments are 2,1, return nth Lucas number, otherwise, return nth Fibonacci number. If other arguments are given, return none until other series can be programmed.
```
PythonCHB, line 39: Ah -- here is the trick -- the goal here is to write this one as a generic function that can compute any series of that form. So write that code, and it should "just work", without needing to actually call your fibonacci or Lucas functions at all.

PR #20, opened 2014-10-15
Homework assignments

agrootsky:


Students/apklock/session02/dist.py
```diff
@@ -0,0 +1,6 @@
+def dist(x1, y1, x2, y2):
+	x3 = abs(x1 - x2)
+	y3 = abs(y1 - y2)
```
PythonCHB, line 3: do you need the abs() ? you are going to square the value later...

Students/apklock/session02/sum_series.py
```diff
@@ -0,0 +1,6 @@
+def sum_series(l,m,n): # return the 'lth' place in the series with starting values of m and n
```
PythonCHB, line 1: you might want to check for m,n < 0.

And you do want to put some test code in here with "assert" statements.

PythonCHB:
Looking good!

PR #22, opened 2014-10-15
upload new files to class folder

ghost:


Students/brydavis/session02/ack.py
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env python
+
+from sys import argv
+
+len_argv = len(argv)
```
PythonCHB, line 5: very slick nicely done.

Students/brydavis/session02/ack.py
```diff
@@ -0,0 +1,33 @@
...
+_m = int(argv[1]) if len_argv > 1 else 4
+_n = int(argv[2]) if len_argv > 2 else 4
+
+def ack(m,n):
+    '''Demonstrate the power of the recursive via the Ackermann function.'''
+    if m == 0:
```
PythonCHB, line 11: what if m or n is less than 0?

Students/brydavis/session02/ack.py
```diff
@@ -0,0 +1,33 @@
...
+    elif m > 0 and n == 0:
+        return ack(m-1,1)
+    elif m > 0 and n > 0:
+        return ack(m-1,ack(m, n-1))
+        
+
+if __name__ == '__main__':
+    # print 'ack.py is running...'
+    '''
+    Test each pair of inputs between 0 and 4 and assert that the result produced by your function is the result expected by the wikipedia table.
```
PythonCHB, line 22: the idea here was to write tests that use an assert to make sure the values are correct -- so you don't need tp look at the results to know it's all working.

Students/brydavis/session02/series.py
```diff
@@ -0,0 +1,89 @@
...
+    equals the one parameter given, 'n'.  
+    The first and second intergers are 0 and 1, respectively. 
+    '''
+    
+    fibo_seq = [0,1]
+    
+    for i in range(2,n):
+        fibo_seq.append(fibo_seq[i-2] + fibo_seq[i-1])
+
+    return fibo_seq
```
PythonCHB, line 18: interesting -- so the whole sequece gets returned -- nice.

Students/brydavis/session03/list_lab.py
```diff
@@ -0,0 +1,146 @@
...
+
+
+
+
+if __name__ == '__main__':
+	# When the script is run, it should run four series of actions:
+	series_one()
+
+
+# Q: use original 'fruit' list for each new series?
```
PythonCHB, line 138: sure, why not?

Students/brydavis/session03/rot13.py
```diff
@@ -0,0 +1,22 @@
...
+
+from sys import argv
+
+s =  argv[1] if len(argv) > 1 else 'Hello World!'
+
+def rot13(s=s):
+	hashed = ''
+
+	for i in s:
+		hashed += chr(ord(i)+13)
```
PythonCHB, line 11: what happens to the letters toward the end of the alphabet? 

you need to have it roll around...

PR #25, opened 2014-10-16
session03 List Lab Homework

jerrytozhang:


Students/Hui Zhang/session03/lab62.py
```diff
@@ -0,0 +1,21 @@
...
+    (Bonus: Multiply the list times two. Keep asking until a match is found. Once found, delete all occurrences.)
+
+"""
+fruit2 = ['Apples', 'Pears', 'Oranges', 'Peaches']
+print fruit2.pop()
+fruit3 = 2*fruit2
+print fruit3
+str2 = raw_input("Please input a fruit name here to be removed: ")
+fruit4 = []
+for i in range(len(fruit3)):
```
PythonCHB, line 18: you should be able to do this without keeping the index:
for f in fruit3:
    if fruit3 != str2:
        append(f)

a bit cleaner....

Students/Hui Zhang/session03/lab63.py
```diff
@@ -0,0 +1,25 @@
...
+Ask the user for input displaying a line like “Do you like apples?”
+ for each fruit in the list (making the fruit all lowercase).
+ For each “no”, delete that fruit from the list.
+ For any answer that is not “yes” or “no”, prompt the user to answer with one of those two values (a while loop is good here):
+ Display the list.
+"""
+
+fruit1 = ['Apples', 'Pears', 'Oranges', 'Peaches', 'Mongo', 'Banana']
+fruit5 = []
+for i in range(len(fruit1)):
```
PythonCHB, line 14: again, beter to loop directly through the items, rather than using the indices:

for f in fruit1:
   etc....

PythonCHB:
Pull request looks good.

Do remember to loop through the item sin a list, rather than the indexes -- you almost always can do that.

PR #26, opened 2014-10-16
Session3 -- in progress

Anastomose:
I've finished the in-class labs and started the homework so it seemed like a good time to check in and get some git practice.


Students/ebuer/session_03/string_funcs.py
```diff
@@ -0,0 +1,41 @@
...
+    return t_str[4:-4:2]
+
+
+#reverse a string using slicing
+def reverso(t_str):
+    return t_str[::-1] #whole string, step/walk backwards
+
+#return middle, last, first thirds of random string
+def scrambler(t_str):
+    temp = len(t_str)
```
PythonCHB, line 29: maybe store len(t_str) / 3 instead? I think it's used that way everywhere...

PR #28, opened 2014-10-17
All assignments done

voite1:
Please, take a look at them.  Any recommendations for improvement?


Students/A.Kramer/session03/ROT13.py
```diff
@@ -0,0 +1,30 @@
+'''
+Created on Oct 1, 2014
+
+@author: db345c
+'''
+
+def rot13(s):
+    alpha1 = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz .!<>;:'"
```
PythonCHB, line 8: interesting approach.

Check out the stri.translate() method -- it does kind of the same thing, but much faster.

or you could use the math -- ord(letter) will give you the numerical value, which you can add 13 to, then use chr() to make it a letter.

but this is fine, too -- more than one way to write a function.

Students/A.Kramer/session03/mailroom.py
```diff
@@ -0,0 +1,73 @@
...
+        else:
+            avg_amnt = donor[1]/donor[2]
+        print "Name: %s\tTotal Donated: $%2.2f\tNumber of Donations: %i\tAverage Donation: $%2.2f" % (donor[0], donor[1], donor[2], avg_amnt)
+    print ""
+    
+def generate_letters(d):
+    """Generate Thank you letter"""
+    name = select_donor(d)
+    donation = get_donation(name)
+    email_text = "Dear " + name + ",\n\nThank you very much for your generous donation of $" + str(donation) + ".\nThis donation will go a long way."
```
PythonCHB, line 35: try tho give string formatting a try for this -- much better if it gets more complicated.

Students/A.Kramer/session03/ROT13.py
voite1, line 8: I did not even think of ord() at the time of writing.  I will try one
tomorrow.

Students/A.Kramer/session03/mailroom.py
voite1, line 35: will try tomorrow.

PR #37, opened 2014-10-19
Session03

QuietMike:
Hello Chris,

I finished (I believe) the mailroom assignment. Would you mind taking a look at it and let me know what you think?

Thanks a lot!

Michel.

PS: Thank you very much for the help with git & github today.

### DIFF
--- a/Students/michel/session03/mailroom.py
+++ b/Students/michel/session03/mailroom.py
@@ -105,12 +105,12 @@ def thankYouPath(donorList, choice):
     """
     while choice == 1:
         print 'Enter "list" to get the list of donors'
-        print 'Enter "3" to exit this activity'
+        print 'Enter "9" to exit this activity'
         name = str(raw_input('Otherwise, please enter a name: '))
         if name == 'list':
             listDonors(donorList)
-        elif name == '3':
-            choice = 3
+        elif name == '9':
+            choice = 9
             break
         else:
             if not checkName(donorList, name):                
@@ -136,13 +136,13 @@ def sortDonorList(donorList):
     return donorList
     
 
-def computeStats(donorList):
+def createReport(donorList):
     """
     Computes the number of donations and average donation per donor
-    Displays the results of the computations
-    returns none
+    Displays the results of the computations in a report
+    returns list of donors
     """
-    sortDonorList(donorList)
+    donorList = sortDonorList(donorList)
     print 'Name', ' ' * 26, 
     print 'Total', ' ' * 5, 
     print 'Donations', ' ' * 7,
@@ -151,15 +151,15 @@ def computeStats(donorList):
     for i in range(len(donorList)):
         name = donorList[i][0][0]
         total = str(donorList[i][2][0])
-        nbDonation = str(len(donorList[i][0][0]))
-        average = str(round((float(total) / float(nbDonation))))
+        nbDonation = str(len(donorList[i][1]))
+        average = str(round((float(total) / float(nbDonation)),2))
         print name, ' ' * (20-len(donorList[i][0][0])),
         print total.rjust(15),
         print nbDonation.rjust(15),
         print average.rjust(15)
-    return None
-        
-    
+    return donorList
+
+
 def selectPath():
     """
     Offers user 3 choices: Write a thank you email, create a report, or quit
@@ -178,3 +178,20 @@ def selectPath():
         else:
             break
     return int(choice)
+    
+    
+if __name__ == '__main__':
+    currentList = []
+    choice = 0
+    currentList = initList(currentList)
+    while choice != 3:
+        choice = selectPath()
+        if choice == 1:
+            currentList, choice = thankYouPath(currentList, choice)  
+        elif choice == 2:
+            currentList = createReport(currentList)
+        elif choice == 3:
+            print 'Bye!'
+            break
+
+        
--- a/Students/michel/session03/mailroom.py
+++ b/Students/michel/session03/mailroom.py
@@ -44,7 +44,7 @@ def thankYouMail(name, amount):
     amount is integer with donation
     """
     emailBody = name + ", we wanted to thank you for your generous donation of "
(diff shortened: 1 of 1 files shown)


Students/michel/session03/mailroom.py
```diff
@@ -0,0 +1,197 @@
...
+"""
+
+def initList(donorList):
+    """
+    Initiatilize list of donors with names and respective donation history
+    returns pre-populated list with 5 donors and their donation activity
+    donorList is a list of donors with donors names, list of amounts donated
+    and the total amount given so far
+    """
+    Andrew = [['Andrew'], [25, 35, 15], [75]]
```
PythonCHB, line 15: why is the name in a list, rather than on its own?

and why the 75 in a list by itself, too?

PythonCHB, line 15: Ah -- I see -- the 75 is the total. but no need for it to be a one-element list, rather than on it's own. Also, be carful about storing values that can be calculated for other data -- yu then need to make sure they don't get out of sync.

Students/michel/session03/mailroom.py
```diff
@@ -0,0 +1,197 @@
...
+    return donorList
+    
+    
+def updateTotals(donorList):
+    """
+    Compute and update total amount given by each donor in the donor list
+    returns list of donor with updated total donation
+    donorList is the list of donors with donations to be updated
+    """
+    for i in range(len(donorList)):
```
PythonCHB, line 32: you rarely want to use indexes to loop through a list. why not:

for donor in donorList:
       total = 0
        for j in (donor[1]):
            total += j
        donor[2][0] = total
    return donorList

Also -- this will update the passed-in list, so  you don't really need to return in -- but in any case, you should certainly document that the list is updated in the docstring.

Students/michel/session03/mailroom.py
```diff
@@ -0,0 +1,197 @@
...
+        total = 0
+        for j in (donorList[i][1]):
```
PythonCHB, line 34: the built-in sum(a_sequence) would be useful here.

Students/michel/session03/mailroom.py
```diff
@@ -0,0 +1,197 @@
...
+    return donorList
+    
+def thankYouMail(name, amount):
+    """
+    Write thank you email to donor name for amount given
+    returns email body with inserted name and amount in a string
+    name is a string with valid name of a donor 
+    amount is integer with donation
+    """
+    emailBody = name + ", we wanted to thank you for your generous donation of "
```
PythonCHB, line 46: give this a shot with string formatting -- either the % operator or the format method. adding string is fine when it's this simple, but formatting is a lot more flexible and faster when things get a bit more complicated.

Students/michel/session03/mailroom.py
```diff
@@ -0,0 +1,197 @@
...
+    emailBody = emailBody + "$" + str(amount) + "\rThank you \rThe Team"
+    return emailBody
+    
+    
+def checkName(donorList, name):
+    """
+    Checks whether a name entered is in the list already 
+    returns True if name is found else returns False
+    """
+    for i in range(len(donorList)):
```
PythonCHB, line 56: again, no need to lop through the indexes:

for donor in donorList:
    if donor[0][0] == name
    .....

or better yet if the name was not in a list:

if the donor name was ont in a list:

if donor[0] == name:
....

Students/michel/session03/mailroom.py
```diff
@@ -0,0 +1,197 @@
...
+            return True
+    return False
+    
+
+def listDonors(donorList):
+    """
+    Prints list of all donors in database
+    returns None
+    """
+    for i in range(len(donorList)):
```
PythonCHB, line 67: sothis could be:

for donor in donrList:
    print donor[0]

or even:

for name, donations, total in donorList:
    print name

(you can unpack multiple values in a for loop, too!)

Students/michel/session03/mailroom.py
```diff
@@ -0,0 +1,197 @@
...
+    
+def addDonor(donorList, name):
+    """
+    Add donor name to donorList
+    returns updated donorList
+    assumes name does not exist already in the list
+    name is a string
+    """
+    newDonor = [[name], [], [0]]
+    donorList = donorList.append(newDonor)
```
PythonCHB, line 80: nice -- though I'd actually probably use a tuple, rather than a list, for the donor data structure. tuples are often used to group realted things with different data types and meanings:

newDonor = (name, [], [0])

(In this case, you do need the total in the list, so you can change it, but I wouldn't store the total, anyway). You can mutate the list of donations, even if they are inside a tuple.

Students/michel/session03/mailroom.py
```diff
@@ -0,0 +1,197 @@
...
+    
+
+def createReport(donorList):
+    """
+    Computes the number of donations and average donation per donor
+    Displays the results of the computations in a report
+    returns list of donors
+    """
+    donorList = sortDonorList(donorList)
+    print 'Name', ' ' * 26,
```
PythonCHB, line 146: see if you can do this with string formatting, too

Students/michel/session03/mailroom.py
```diff
@@ -0,0 +1,197 @@
...
+    return donorList
+
+
+def selectPath():
+    """
+    Offers user 3 choices: Write a thank you email, create a report, or quit
+    returns choice
+    choice is an integer between 1-3
+    """
+    choice = ''
```
PythonCHB, line 169: no need to initialize choice to an empty string.

Students/michel/session03/mailroom.py
```diff
@@ -0,0 +1,197 @@
...
+    print 'Please, select an option: '
+    print '1- Send a thank you email'
+    print '2- Create a donation report'
+    print '3- Quit the application'
+    while True:
+        choice = str(raw_input('Your Selection: '))
+        if choice not in ['1','2','3']:
```
PythonCHB, line 176: nice!

PR #38, opened 2014-10-19
Session3 -- mailroom check-in

Anastomose:
Nearly done with mailroom.py.  Still need to add final if **name** ==' **main**' block but the other components are finished so it seemed like a good time to check in.

### DIFF
--- a/Students/ebuer/session_03/mailroom.py
+++ b/Students/ebuer/session_03/mailroom.py
@@ -32,10 +32,10 @@ def name_split(nme):
 def client_check(nme, donors):
     fname, lname = nme.split(' ')
     if (lname, fname) in donors:
-        print '{f:s} {l} is a previous donor.'.format(f=fname, l=lname)
+        print '{f} {l} is a previous donor.'.format(f=fname, l=lname)
         return True
     else:
-        print '{f:s} {l} is a not a previous donor.'.format(f=fname, l=lname)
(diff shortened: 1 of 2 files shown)


Students/ebuer/session_03/mailroom.py
```diff
@@ -1,33 +1,166 @@
 #mailroom program to break the monotony
 
-import random
+#this may be a bad idea but we will see
+#a list of tuples for donors and donations
```
PythonCHB, line 4: nice -- I like the separation of first and last names.

However, you may not want to use a tuple to store the donations, as you'll want to add to those -- so a mutable list is easier.

Anastomose, line 4: Agreed, it wasn't the best solution but I wasn't sure how else I could ensure the names and donations remained firmly linked.  In the end this scheme required creating a new record with the added donation and replacing the entire existing record (name + donation tuple) for each loop.

PythonCHB, line 4: if you make each reocrd a tuple, it will keep the name and donation list linked, but the dontation list can still be mutable -- i.e. a list:

donor_record = ( 'A name', [34, 67,] )

no you can't change the name or put a different list in the record, but you Can add something to the list.

Anastomose:
Ah, I see!  Thanks for the pointer.  Your data structure is much better.  That would have saved me a bit of work downstream.  Well ... there's always next time.

PR #40, opened 2014-10-20
Session03 assignments

imdavis:
Added scripts for the list lab, rot13, and mailroom assignments for this week.

### DIFF
--- a/Students/imdavis/session03/list_lab.py
+++ b/Students/imdavis/session03/list_lab.py
@@ -1,0 +1,61 @@
+#!/usr/bin/env python2.7
+fruits = ["Apples", "Pears", "Oranges", "Peaches"]
+
+print "List of fruit: ", fruits
+
+fruits.append(raw_input("Add another fruit: "))
+print "Added your fruit to the list: ", fruits
+
+fruit_num = int(raw_input("Give the number of your favorite fruit: "))
+
+if(fruit_num <= 0 or fruit_num > len(fruits)):
+    print "The fruit # you entered: ", fruit_num, "you entered is not in the list!"
+else:
+    print "Your favorite fruit is #:", fruit_num, ":", fruits[fruit_num - 1]
+
+fruits = [raw_input("Add a fruit to the beginning of the list: ")] + fruits
+print "Added your new fruit to the front of the list: ", fruits
+
+fruits.insert(0, raw_input("Add another fruit to the beginning of the list in a different way: "))
+print "Added your new fruit to the front of the list: ", fruits
+
+print "All the fruits in the list that begin with 'P':"
+aPfruit = False
+for afruit in fruits:
+    if ("P" in afruit):
+        aPfruit = True
+        print afruit
+if (not aPfruit):
+    print "None of the fruits start with 'P'"
+
+print "Here is the current list of fruits:", fruits
+print "Removing the last one from the list:", fruits.pop()
+print fruits
+
+fruits.remove(raw_input("What fruit would you like to delete from the list: "))
+print fruits
+
+# iterate over a copy, hence the "fruits[:]:"
+fruits = ["Apples", "Pears", "Oranges", "Peaches"]
+for fruit in fruits[:]:
+    likefruit = ""
+    afruit = fruit.lower()
+    while (likefruit != "yes" and likefruit != "no"):
+        likefruit = raw_input("Do you like " + afruit + "?: ")
+        likefruit = likefruit.lower()
+        if(likefruit != "yes" and likefruit != "no"):
+            print "You entered: ", likefruit, "which is not valid"
+            print "Please enter 'yes' or 'no'"
+    if(likefruit == "no"):
+        fruits.remove(fruit)
+print fruits
+
+# Make a shallow copy, and reverse the letters of each fruit in the copy
+fruits = ["Apples", "Pears", "Oranges", "Peaches"]
+fruits_copy = fruits[:]
+for i, fruit in enumerate(fruits_copy):
+    fruits_copy[i] = fruit[::-1]
+
+# Delete the last item of the original list, and display the list and the copy
+fruits.pop()
+print fruits_copy, fruits
--- a/Students/imdavis/session03/mailroom/mailroom.py
+++ b/Students/imdavis/session03/mailroom/mailroom.py
@@ -1,0 +1,40 @@
+#!/usr/bin/env python2.7
+
+from mailroomfunct import *
+
+# Hardcoded original group of donors and donation amounts
+donors = []
+donors.append(("Robert Plant", [15.00, 25.32, 100.50]))
+donors.append(("Sandra Bullock", [12.50, 2.25]))
+donors.append(("Richard D. James", [1500.34, 2349.99]))
+donors.append(("Slash", [1.00, 10.99]))
+donors.append(("Jessica Alba", [13.49]))
+
+# Initial greeting at startup
+print "Welcome to Mailrooom, buddy!"
+
+# Call the prompt function to ask the user what they would like to do
+todo = ""
+# keep looping with prompts until the user says to exit
+while (todo != "exit"):
+    # send a thank you block
+    todo = prompt1()
+    if(todo == "send a thank you"):
+        # get the index of existing or new donor
+        indexofdonor = donorindex(donors)
+        # get the donation amount
+        donation = newdonation(donors[indexofdonor][0])
+        # add the new donation amount to the appropriate donor
+        donors[indexofdonor][1].append(donation)
+        # update the user of who the donor is and their updated donation history
+        print "Donor:", donors[indexofdonor][0]
+        print "Donation History:", donors[indexofdonor][1]
+        # compose the email message thanking the donor for their recent donation
+        composemail(donors[indexofdonor][0], donors[indexofdonor][1][-1])
+    # create a report block
+    elif(todo == "create a report"):
+        # print a header for the report table
+        formattable("Donor Name", "Total Donations", "Number of Donations", "Average Donation ($)")
+        # print each row of the table for each donor
+        for i in range(len(donors)):
+            print_donor_row(donors[i][0], donors[i][1])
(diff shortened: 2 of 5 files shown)


Students/imdavis/session03/list_lab.py
```diff
@@ -0,0 +1,61 @@
...
+
+if(fruit_num <= 0 or fruit_num > len(fruits)):
+    print "The fruit # you entered: ", fruit_num, "you entered is not in the list!"
+else:
+    print "Your favorite fruit is #:", fruit_num, ":", fruits[fruit_num - 1]
+
+fruits = [raw_input("Add a fruit to the beginning of the list: ")] + fruits
+print "Added your new fruit to the front of the list: ", fruits
+
+fruits.insert(0, raw_input("Add another fruit to the beginning of the list in a different way: "))
```
PythonCHB, line 19: slick on-liner!

Students/imdavis/session03/list_lab.py
```diff
@@ -0,0 +1,61 @@
...
+print "Added your new fruit to the front of the list: ", fruits
+
+print "All the fruits in the list that begin with 'P':"
+aPfruit = False
+for afruit in fruits:
+    if ("P" in afruit):
```
PythonCHB, line 25: this would give you a result for APPLES -- do you want that?

checkout str.startswith()

or

if afruit[0] == 'P'

or maybe:

if afruit[0].lower() == 'p'

....

Students/imdavis/session03/list_lab.py
```diff
@@ -0,0 +1,61 @@
...
+    print "None of the fruits start with 'P'"
+
+print "Here is the current list of fruits:", fruits
+print "Removing the last one from the list:", fruits.pop()
+print fruits
+
+fruits.remove(raw_input("What fruit would you like to delete from the list: "))
+print fruits
+
+# iterate over a copy, hence the "fruits[:]:"
```
PythonCHB, line 38: bingo!

Students/imdavis/session03/list_lab.py
```diff
@@ -0,0 +1,61 @@
...
+fruits = ["Apples", "Pears", "Oranges", "Peaches"]
+for fruit in fruits[:]:
+    likefruit = ""
+    afruit = fruit.lower()
+    while (likefruit != "yes" and likefruit != "no"):
```
PythonCHB, line 43: you could get fancy here:

while likefruit not in ('yes, 'no'):

Students/imdavis/session03/list_lab.py
```diff
@@ -0,0 +1,61 @@
...
+        if(likefruit != "yes" and likefruit != "no"):
+            print "You entered: ", likefruit, "which is not valid"
+            print "Please enter 'yes' or 'no'"
+    if(likefruit == "no"):
+        fruits.remove(fruit)
+print fruits
+
+# Make a shallow copy, and reverse the letters of each fruit in the copy
+fruits = ["Apples", "Pears", "Oranges", "Peaches"]
+fruits_copy = fruits[:]
```
PythonCHB, line 55: in this case, you care going to replace all the itmes, so you might as well start empty:

fruits_copy = []

then do append in the loop.

Students/imdavis/session03/mailroom/mailroom.py
```diff
@@ -0,0 +1,40 @@
+#!/usr/bin/env python2.7
+
+from mailroomfunct import *
```
PythonCHB, line 3: careful -- "import *" is almost never a good idea!

PythonCHB, line 3: but nice separation of donor-list manipulating functions to a separate module -- this would make it "easy" to replace the list with persistant storage, for instance -- a DB or ???

Though you probably want to put the Hard coded list in the other module too 00 ideally this module would not need to know what the data structure was.

Students/imdavis/session03/mailroom/mailroom.py
```diff
@@ -0,0 +1,40 @@
...
+        print "Donor:", donors[indexofdonor][0]
+        print "Donation History:", donors[indexofdonor][1]
+        # compose the email message thanking the donor for their recent donation
+        composemail(donors[indexofdonor][0], donors[indexofdonor][1][-1])
+    # create a report block
+    elif(todo == "create a report"):
+        # print a header for the report table
+        formattable("Donor Name", "Total Donations", "Number of Donations", "Average Donation ($)")
+        # print each row of the table for each donor
+        for i in range(len(donors)):
```
PythonCHB, line 39: I'd probably put the entire report printing in a function in the other module. Again, the rule of thumb here is that anything that needs to know the data structure used goes together. So this module handles the user interaction, and the other one handles all the manipulation of the data.

We'll see this idea later with object-oriented design.

Students/imdavis/session03/mailroom/mailroomfunct.py
```diff
@@ -0,0 +1,138 @@
...
+        elif(request == 'e' or request == action3):
+            print "Exiting"
+            return action3
+        else:
+            # if the user doesn't enter an available action, let them know where they went astray.
+            print "You entered: '", orig_request, "'"
+            print "Please enter '(S)end a thank you' or '(C)reate a report', or '(E)xit'."
+
+def finddonorindex(donors, adonor):
+    """
```
PythonCHB, line 41: I like this.

Students/imdavis/session03/mailroom/mailroomfunct.py
```diff
@@ -0,0 +1,138 @@
...
+            for name,donations in donors:
+                print name
+        else:
+            index = finddonorindex(donors, orig_request)
+    return index
+
+def is_number(s):
+    """
+    Couldn't get str.isnumeric working with unicode. This function checks to see
+    if a string will convert to float ok.
```
PythonCHB, line 84: yup -- this is the way to do it -- I"ll talk about that when I introduce Exceptions.

though I don't usually bother pitting it in a function.

Students/imdavis/session03/mailroom/mailroomfunct.py
```diff
@@ -0,0 +1,138 @@
...
+
+    This makes use of the 'is_number' function above.
+    """
+    validamount = False
+    message = "Enter a new donaton amount for donor '" + donorname + "' > "
+    while (not validamount):
+        amount = raw_input(message)
+        validamount = is_number(amount)
+        if (validamount):
+            return float(amount)
```
PythonCHB, line 105: maybe have validamount() return the float? why do it twice?

Students/imdavis/session03/mailroom/mailroomfunct.py
```diff
@@ -0,0 +1,138 @@
...
+        else:
+            print "You entered: '", amount, "' which is not a valid donation amount."
```
PythonCHB, line 107: try using string formatting for this kind of thing

Students/imdavis/session03/mailroom/mailroomfunct.py
```diff
@@ -0,0 +1,138 @@
...
+
+def composemail(donorname, recentdonation):
+    """ 
+    Compose a message to the donor thanking them and rounding the donation amount to the penny.
+    Uses string formatting.
+    """
+    message = "\n\nDear {donor},\nThank you very much for your most recent donation of ${donation:.2f}\n".format(donor=donorname, donation=recentdonation)
+    message +="Please consider donating to our charity again in the future.\nSincerely,\nIan Davis\n\n"
```
PythonCHB, line 115: whynot put this part in the original string? no need to add it later.

Students/imdavis/session03/mailroom/mailroomfunct.py
```diff
@@ -0,0 +1,138 @@
...
+            return float(amount)
+        else:
+            print "You entered: '", amount, "' which is not a valid donation amount."
+
+def composemail(donorname, recentdonation):
+    """ 
+    Compose a message to the donor thanking them and rounding the donation amount to the penny.
+    Uses string formatting.
+    """
+    message = "\n\nDear {donor},\nThank you very much for your most recent donation of ${donation:.2f}\n".format(donor=donorname, donation=recentdonation)
```
PythonCHB, line 114: you also might want to use a tripple quoted string, so as not to have to keep track of all the \n s...

Students/imdavis/session03/mailroom/mailroomfunct.py
```diff
@@ -0,0 +1,138 @@
...
+    message +="Please consider donating to our charity again in the future.\nSincerely,\nIan Davis\n\n"
+    print message
+
+def sumdonations(donationlist):
```
PythonCHB, line 118: this one is built in! try:

sum(donation_list)

Students/imdavis/session03/mailroom/mailroomfunct.py
```diff
@@ -0,0 +1,138 @@
...
+    Simple function to sum donation amounts in a list past to it.
+    """
+    sum = 0.0
+    for amount in donationlist:
+        sum += amount
+    return sum
+
+def formattable(donor, total, ndonations, ave):
+    print "{donor:^20}{total:^20}{ndonations:^20}{ave:^20}".format(donor=donor, total=total, ndonations=ndonations, ave=ave)
+
```
PythonCHB, line 129: I"d probably have this return the string -- then print it when you call it.

Or, as this is a one liner, just put it in where you need it...

Students/imdavis/session03/rot13/translate.py
```diff
@@ -0,0 +1,70 @@
...
+    decrypt using these functions, since there are 26 letters in the alphabet.
+    This is designed to leave whitespace, punctuation, and capitalization 
+    intact in the encrypted string as in the original.
+
+    Lowercase a-z are ordinals 197 -122 (enclusive)
+    For example: ord('s') = 115 and chr(115) = 's'
+
+"""
+
+def lowercase(substring):
```
PythonCHB, line 18: slick idea -- though not very efficient :-(

PythonCHB:
See my notes for little things you could clean up -- but nice work!

PR #42, opened 2014-10-21
Session 3 Homework

salimhamed:
Session 3 Homework for Salim Hamed.


Students/salim/session03/mailroom.py
```diff
@@ -0,0 +1,136 @@
...
+    s = ''
+    for doner in d_list:
+        s += (doner[0] + '\n')
+    return s.rstrip()
+
+
+def in_list(name, d_list):
+    return_bool = False
+    for item in d_list:
+        if type(item[0]) == str:
```
PythonCHB, line 24: no need to check type here -- you are controlling the content of the d_list, and if there is something wrong with, you would want to get an error!

Students/salim/session03/mailroom.py
```diff
@@ -0,0 +1,136 @@
...
+    response_level = 1
+
+    # level 1
+    while response_level >= 1:
+
+        if response_level == 0:
+            break
+
+        elif response_level == 1:
+
```
PythonCHB, line 68: I'd put these in separate functions, so as to keep the code a bit cleaner, but all the logic looks good.

Students/salim/session03/mailroom.py
```diff
@@ -0,0 +1,136 @@
...
+                print display_report(donation_list)
+                response_level = 1
+                response = ''
+
+        elif response_level == 3:
+
+            response = raw_input('Enter donation amount: ')
+
+            # prompt user for correct entry
+            while (not response.isdigit()) and (response not in ['quit', 'back']):
```
PythonCHB, line 122: now that we covered Exceptions, you can make this somethign like:

try:
    float(response)
except ValueError:
   ....

Students/salim/session03/rot13.py
```diff
@@ -0,0 +1,36 @@
...
+                return_s += encrypt(letter, 65)
+
+        # do not encrypt non-alphabet characters
+        else:
+            return_s += str(letter)
+
+    return return_s
+
+
+def encrypt(letter, start):
```
PythonCHB, line 24: looks good

Students/salim/session03/string_lab.py
```diff
@@ -0,0 +1,19 @@
...
+numbers = [1, 4, 83, 321, -56, 88]
+print 'The first 3 numbers are: %i, %i, %i' % tuple(numbers[:3])
+
+# use format()
+
+# automatic field numbering
+print 'file_{:=03} :   {:10.2f}, {:10.0e}'.format(2, 123.4567, 10000)
+
+# positional field numbering
+print 'file_{0:=03} :   {1:10.2f}, {2:10.0e}'.format(2, 123.4567, 10000)
```
PythonCHB, line 13: nice!

PR #43, opened 2014-10-21
add list_lab 

gosunfish:


.idea/dictionaries/carolyn_evans.xml
```diff
@@ -0,0 +1,7 @@
+<component name="ProjectDictionaryState">
```
PythonCHB, line 1: oops -- this shouldn't be in the repo.

you should be able to do:

git rm --cached .idea/dictionaries/carolyn_evans.xml

the --cached keeps git from remving the file itself from your disk.

Then you may want to add .idea to your .git ignore file. It should be in the IntroToPython dir, and should look ike this when your done:

slides_sources/build
.idea

I _think that will tell git not to ever add the .idea files...

Students/CarolynEvans/session03/slicing_lab.py
```diff
@@ -30,7 +30,7 @@ def Slice(seq, task):
         seq = seq[onethird:onethird*2] + seq[0:onethird] + seq[(onethird + remainder) * -1:]
 
     elif task == 'removefirstlastfoureveryother':
-        seq = Slice(seq[4:-4], 'removeeveryother')
+        seq = seq[4:-4:2]
```
PythonCHB, line 33: nice and clean, eh?

PR #44, opened 2014-10-22
Adding session 03 homework, still working on mailroom:wq!

PrimusPilus:
Adding Session03 Homework


PythonCHB:
we'll wait for the next version to final review

PR #45, opened 2014-10-22
adding mailroom.py

jerrytozhang:


Students/Hui Zhang/session03/mailroom.py
```diff
@@ -0,0 +1,14 @@
+from mailroomfunction import sendletter
+from mailroomfunction import printout
+
+Userlist = ['Jerry', [10, 20, 30], 'Peter', [5, 6], 'Aly', [3, 4], 'Anisha', [11, 15, 13], 'Chris', [9, 11, 14], 'Vini', [13, 17, 16]]
```
PythonCHB, line 4: This will work, but probably better to explicitly group the person and the donation list, as I talked about in class:

UserList = [ ( 'Jerry', [10, 20, 20] ), ....

That way you don't need to do a bunch of index manipulations to keep track.

So maybe update with a dict for the next week.

Students/Hui Zhang/session03/mailroom.py
```diff
@@ -0,0 +1,14 @@
...
+
+input1 = input("To send a Thanks You letter, press '1'\nTo create a report, press '2'\nTo Exit, press '0' \n")
+
+while input1:
+    if input1 == 1:
+        sendletter(Userlist, input1)
+        input1 = input("To send a Thanks You letter, press '1'\nTo create a report, press '2'\nTo Exit, press '0' \n")
```
PythonCHB, line 11: you want to try to structure your while loops so you don't need to repeat the input message.usually a while True: at teh top, then the input message, and break out when you want to end it.

Students/Hui Zhang/session03/mailroomfunction.py
```diff
@@ -0,0 +1,55 @@
+
+
+def sendletter(Userlist, input1):
+    # Print out all existing donators name.
+    print Userlist[::2]
```
PythonCHB, line 5: this is a slick way to get the names -- isn't slicing handy?

Students/Hui Zhang/session03/mailroomfunction.py
```diff
@@ -0,0 +1,55 @@
...
+    if __name__ == "__main__":
+        # execute only if run as a script
+        main()
+
+def printout(Userlist, input1):
+    #initilize a list named donator_tuples
+    donator_tuples = []
+    # create a list wiht donator name, sum of donation, times of donation and average donation.
+    for username2 in Userlist[::2]:
+        locationuser2 = Userlist.index(username2)
```
PythonCHB, line 35: this is where grouping the user and donations would help. This works pretty well, but is not very efficient, the code is searching through the whole user list multiple times. Not a big deal with a small data set, but would kill you with a big one.

Students/Hui Zhang/session03/mailroomfunction.py
```diff
@@ -0,0 +1,55 @@
...
+        amount2 = sum(Userlist[locationuser2 + 1])
+        # calculate how many times of donation
+        length2 = len(Userlist[locationuser2 + 1])
+        # calculate average donation
+        average2 = int(amount2 / length2)
+        donator_tuples.append([username2, amount2, length2, average2])
+    # sort the list: donator_tuples by sum of donation.
+    donator_tuples_sorted = sorted(donator_tuples, key=lambda donator: donator[1])
```
PythonCHB, line 43: nice use of sorted and lambda!

Students/Hui Zhang/session03/mailroomfunction.py
```diff
@@ -0,0 +1,55 @@
...
+    print '\n\n'
+    # Left align strings to be printed out
+    print '{0:<10} {1:<8} {2:<8} {3:<8}'.format("Donator", "Total", "Times", "Average")
```
PythonCHB, line 46: very nice formatting

PR #47, opened 2014-10-24
Homework 4

voite1:
Please, look at homework 4 assignments.


Students/A.Kramer/session03/ROT13_3.py
```diff
@@ -0,0 +1,21 @@
...
+'''
+import string
+
+def rot13(s):
+    dt = string.maketrans( 
+    "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz .!<>;:'", 
+    "NOPQRSTUVWXYZabcdefgjijklmnopqrstuvwxyzABCDEFGHIJKLM .!<>;:'")
+    
+    return string.translate("Hello World!", dt)
+
```
codedragon, line 14: You are not actually using your input here. try: 
return s.translate(dt)

Students/A.Kramer/session04/CurrentPath.py
```diff
@@ -0,0 +1,18 @@
+'''
+Created on Oct 23, 2014
+
+@author: db345c
+'''
+import os
+
+def printCurrentDirectory(dir1=os.getcwd()):
+    """ Print the content of the directory """
```
codedragon, line 9: os.getcwd() is going to be called at the time that this function is created. So, if you change directories after you import this function, it is going to give you the contents of the original directory, not your current directory. Be very careful with functions as input defaults, and only use them if you know you want the input to be the result at the time of creation only.

Students/A.Kramer/session04/FileCopy.py
```diff
@@ -0,0 +1,31 @@
...
+        # The actual copy
+        for line in frm:
+            to.write(line)
+        print "Copy successful"
+        try: 
+            frm.close()
+            to.close()
+        except IOError:
+            print "Error closing either source or dest. files"
+
```
codedragon, line 25: This works, but you can also copy the entire contents at once, instead of line by line. See Chris's solution.

Students/A.Kramer/session04/kakta14.py
```diff
@@ -0,0 +1,65 @@
...
+    count = 0
+    for line in f:
+        # Split the line into separate tokens
+        lst = line.split()
+        
+        # discard all the lines containing less than three words
+        if len(lst) > 2:
+            
+            # Remove punctuation with string.punctuation (import string)
+            for j, k in enumerate(lst):
```
codedragon, line 25: Fun place for a list comprehension:
lst = [i.strip(string.punctuation) for i in lst]

Students/A.Kramer/session04/kakta14.py
```diff
@@ -0,0 +1,65 @@
...
+            lst_length = len(lst)
+            for idx, i in enumerate(lst):
+                # assure the last element of the list is reached, if so, break
+                if idx == lst_length - 2:
+                    break
+                else:
+                    # got the three strings to use and add those to the dictionary
+                    l = [ (lst[idx] + " " + lst[idx + 1]).strip(), lst[idx + 2].strip()]
+                    buildDictionary(l, words)
+                    count += 1
```
codedragon, line 38: count doesn't do anything

Students/A.Kramer/session04/kakta14.py
```diff
@@ -0,0 +1,65 @@
...
+        if len(word[start]) > 0:
+            temp = word[start].pop()
+            print start, temp,
+            printKata14(word, str(start.split()[1]) + " " + temp)
+        else:
+            return
+        
+if __name__ == "__main__":
+    """ Execute Kata14 algorythm """
+    kata14("sherlock.txt", "little use")
```
codedragon, line 65: Think about how you can use all of the words from the text (currently not using last couple of words of a line or words at the end of a paragraph if less than 3 words on the last line). Good work, this is a difficult one.

Students/A.Kramer/session04/lecture_notes.py
```diff
@@ -0,0 +1,97 @@
...
+'''
+
+def inClass():
+      
+    l = range(10)
+    
+    for i in range(len(l)):
+        print i
+    print
+
```
codedragon, line 14: This was meant as an example of not very pythonic code. Rarely you need len(my_variable) in a for loop. Can use 'for i in my_variable' or 'for i in range(10)' usually.

Students/A.Kramer/session04/kakta14.py
voite1, line 65: That is an easy one.  Just accumulate the leftover words on a separate
string and process it last.

A. K.

codedragon, line 65: That technically works, but it kind of goes against the grain of the idea behind the assignment. The idea is that a particular author sounds a certain way, because they tend to use words in a certain order, so you can make something random that sort of sounds like the author by mixing the words up, but not too much. This is why when you create your 'random' text based on their writing, you follow two words they use together with a third word that they used (at least once) to follow the first two. If you make a big catch-all of the leftover words, this ordering gets lost. To preserve the intent of the assignment, you would make the entire text into one big list of words, and then put it into the dictionary: the first two words linked to the third word, the second and third word linked to the fourth word, and so on, for the entire text.

PR #48, opened 2014-10-25
Finished Week 3 Homework and Lab

saschwafel:
Hey Chris, I know this is super late but I was completely stumped with mailroom.py

Thanks for any feeback!

### DIFF
--- a/Students/SSchwafel/session03/mailroom.py
+++ b/Students/SSchwafel/session03/mailroom.py
@@ -10,24 +10,21 @@ def historical_amounts(x):
     while full_name.isdigit() == False :
 
         full_name = raw_input('Please enter the full name of your donor:\n')
-        full_name = full_name.title()
+        full_name = full_name.lower()
         
-        print donors
-            
         if is_name_present(full_name, donors) == True:
-            print 'baller alert'
 
+            print 'It looks like in the past, ' + full_name.title() + ' donated: ' + str(donation_amount(full_name,donors)).strip("['']")
+
+            query_new_donation = raw_input('Would you like to add a new donation for this donor? \n\n(yes/no)\n\n')
+
+            if query_new_donation.lower() == 'yes':
+                add_donation(full_name)
         else: 
 
             print 'Keep hackin'
 
-            #print full_name == i[0]
 
-
-#            #index_number = int(x.index(full_name)) + 1
-#
-#            print 'It looks like in the past, ' + full_name + ' donated: '# +  str(x[index_number])
-#
 #            query_new_donation = raw_input('Would you like to add a new donation for this donor? \n\n(yes/no)\n\n')
 #                
 #            if query_new_donation.lower() == 'yes':
@@ -68,14 +65,23 @@ def historical_amounts(x):
 
 def is_name_present(x,y):
 
-        for i in y:
-            #figure out if donor is in the donors list
+    for i in y:
 
-            if x in i:
-                return True
-            else: 
-                return False    
+        if x == i[0]:
+            return True
+        
+    return False     
+            
+def donation_amount(x,y):
+    """Queries donation amount for donor x"""
+    #x = full_name
+    #y = donors
 
+    for i in y:
+
+        if x == i[0]:
+            return i[1]
+        
 def add_donation():
 
     new_donation = raw_input('What is the latest donation amount from this donor? ')
@@ -108,8 +114,8 @@ user_input = user_input.title()
 
 if user_input.upper() == 'A':
     #do something
-    #print historical_amounts(donors)
-    print is_name_present('Hal Jordan', donors)
+    print historical_amounts(donors)
+    #print is_name_present('Hal Jordan', donors)
 
 #elif user_input == 'Create A Report':
 #
--- a/Students/SSchwafel/session03/mailroom.py
+++ b/Students/SSchwafel/session03/mailroom.py
@@ -23,7 +23,6 @@ def send_a_thank_you(x):
             if query_new_donation.lower() == 'yes':
 
                 add_donation(full_name)
-                print donors
 
             elif query_new_donation.lower() == 'no':
 
@@ -72,22 +71,22 @@ def create_a_report():
             #donor_name = i[0]
             #sorting_key = i[1]
 
-            temp_list.append((i[0],len(i[1]),i[1]))
+            total_amount_of_donations = (sum(i[1])/len(i[1]))
 
-            # len(i[1]])
+            temp_list.append((i[0],len(i[1]),i[1]))
+            
             #temp_list.append(str())
 
 
     #sorted_donors = '\n'.join(map(str, sorted_donors))
-    temp_list = '\n'.join(map(str, temp_list))
+
+    #temp_list = '\n'.join(map(str, temp_list))
 
     for i in temp_list:
 
-        for x in i:
-    
-            print "{0}".format(i)
+        print '%s %s %s' % (i[0],i[1],i[2])
 
-    #print temp_list 
+    print temp_list 
 
     #sorted_donors =  str(sorted_donors)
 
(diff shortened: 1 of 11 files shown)


Students/SSchwafel/session03/mailroom.py
```diff
@@ -1,112 +1,177 @@
 #!/usr/bin/python
 
-donors = ['bruce wayne',[100000], 'clark kent',[50000], 'barry allen',['100'], 'diana prince',['1300'], 'wally west',['10034'],'hal jordan',['10000']]
 
-def historical_amounts(x):
+donors = [('bruce wayne',[100000]), ('clark kent',[50000]), ('barry allen',[100]), ('diana prince',[1300]), ('wally west',[10034]),('hal jordan',[10000])]
```
PythonCHB, line 4: nice -- definitely better to group the related data in a tuple.

Students/SSchwafel/session03/mailroom.py
```diff
@@ -1,112 +1,177 @@
...
-
-        if full_name in x:
-            #figure out if donor is in the donors list
-            #if so, print the index AFTER their name (the list of their donations)
-
-            index_number = int(x.index(full_name)) + 1
+        full_name = raw_input("Please enter the full name of your donor (you can type 'back' to return to the previous prompt):\n\n")
+        full_name = full_name.lower()
+        
+        if is_name_present(full_name, donors) == True:
```
PythonCHB, line 17: in general, you want to use:

if something is True

rather than == True.

also for False and None.

Students/SSchwafel/session03/mailroom.py
```diff
@@ -1,112 +1,177 @@
...
 
-            print 'It looks like in the past, ' + full_name + ' donated: ' +  str(x[index_number])
+            print 'It looks like in the past, ' + full_name.title() + ' donated: $' + str(donation_amount(full_name,donors)).strip("[]")
```
PythonCHB, line 19: In the future, try to so with is sting formatting -- somethign like:
(untested)
'It looks like in the past, {} donated: ${}'.format(full_name.title(), donation_amount(full_name,donors))

Students/SSchwafel/session03/mailroom.py
```diff
@@ -1,112 +1,177 @@
...
+donors = [('bruce wayne',[100000]), ('clark kent',[50000]), ('barry allen',[100]), ('diana prince',[1300]), ('wally west',[10034]),('hal jordan',[10000])] 
+
+def send_a_thank_you(x):
 
     """Displays a list of all the donations a user has contributed"""
 
     full_name = ''
-    
 
     while full_name.isdigit() == False :
```
PythonCHB, line 12: why are you checking for isdigit() here?

Students/SSchwafel/session03/mailroom.py
```diff
@@ -1,112 +1,177 @@
...
 
+            if query_add_donor.lower() == 'yes':
+                donors.append(add_donor(full_name))
+                break
+
+            elif query_add_donor.lower() == 'no':
+                break
             else:
-                
-                #changes full_name to a digit value, killing the while loop
```
PythonCHB, line 59: I see now -- that's how you were braking the loop -- you're right, break is better -- but then the loop could be:

while True:

Students/SSchwafel/session03/mailroom.py
```diff
@@ -1,112 +1,177 @@
...
-                full_name = str(1)
+                break
+
+
+def create_a_report():
+
+    temp_list = []
+
+    sorted_donors = sorted(donors, key=lambda tup: tup[1])
```
PythonCHB, line 65: nice!

Students/SSchwafel/session03/mailroom.py
```diff
@@ -1,112 +1,177 @@
...
+
+    print '\n{:<12}{:^12}{:>12}'.format('Donor Name','Donations','Total Donations')
+
+    for i in sorted_donors:
```
PythonCHB, line 69: I"d tend to use a more meaningful name here, like 'donor' -- is is usally an index or small integer value...

Students/SSchwafel/session03/mailroom.py
```diff
@@ -1,112 +1,177 @@
...
-    new_donation = raw_input('What is the latest donation amount from this donor? ')
-    return new_donation
-    
+    return (donor_name,[new_donor_initial_donation])
 
-def show_donors():
-    """Returns a list of the donors in organized format"""
-    #print '\n'.join(donors)
-    for i in donors:
+def is_name_present(x,y):
```
PythonCHB, line 88: probably better to use meaningful variable names-- something like:

name, donor_list

so later int eh code it's clear what's going on.

Students/SSchwafel/session03/mailroom.py
```diff
@@ -1,112 +1,177 @@
...
+
+        if x == i[0]:
+            return i[1]
+        
+def add_donation(x, y):
+    """Adds donation to nested list associated with donor"""
+
+    new_donation = raw_input("What is the latest donation amount from this donor? ")
+
+    if new_donation.isdigit() == False:
```
PythonCHB, line 112: this is a good palce to use excetpions:

try:
   new_donation = float(new_donation) # or int
except ValueError:
     return 'Please try again with an int'

Students/SSchwafel/session03/mailroom.py
```diff
@@ -1,112 +1,177 @@
...
+    """Sends prints a message to standard out thanking the user for their donation and summing their donation history. """
+    #x = the list element of our donor
+
+    donation = donation_amount(x, donors)
+
+    return "\n\n\nHello %s,\n\nIt looks like in the past you donated $%i to our organization!\n\nThank you for your contribution, it's people like you who are making a difference.\n\nThank You,\n\nSchuyler INC. Staff\n\n\n" %(x.title(), int(sum(donation)))
 
-user_input = raw_input(
+count = 1
+while count > 0:
```
PythonCHB, line 137: usually better to have this be conditional on something that is naturally changing, or to do while True, and then call break in the loop when you want to break out.

Students/SSchwafel/session04/all_languages.py
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/python
+
+file = open('students.txt').readlines()
```
PythonCHB, line 3: file is  built-in, so generally better not to use it as a vaiable.

why not just:
students_language = open('students.txt').readlines()

Students/SSchwafel/session04/copy.py
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/python
+
+import os
+
+dir = os.getcwd()
+print dir
+
+infile = open('testfile.txt')
```
PythonCHB, line 8: careful, -- you probably want to open this in binary mode:

open('testfile.txt', 'rb')

(same for writing)

Students/SSchwafel/session04/dictionarylab.py
```diff
@@ -0,0 +1,109 @@
...
+#numbers = dict(numbers)
+hexnumbers = []
+
+for i in numbers:
+
+    hexnumbers.append(hex(i))
+
+zipped_dict = zip(numbers, hexnumbers)
+
+print zipped_dict
```
PythonCHB, line 44: this is now a list of tuples -- you can make a dict out ot it with:

dict(zip(numbers, hexnumbers))

Students/SSchwafel/session04/dictionarylab.py
```diff
@@ -0,0 +1,109 @@
...
+p.add('i')
+
+print p
+
+#Create a frozenset with the letters in ‘marathon’
+m = frozenset('marathon')
+print m
+#display the union and intersection of the two sets.
+print p.union(m)
+print p.intersection(m)
```
PythonCHB, line 109: all looks good.

Students/SSchwafel/session04/kata14.py
```diff
@@ -0,0 +1,80 @@
...
+
+#Divide text_doc into a list
+
+text_doc = text_doc.replace('--',' ')
+
+text_doc = text_doc.replace('.','')
+text_doc = text_doc.replace(',','')
+text_doc = text_doc.lower()
+
+text_doc = list(text_doc.split())
```
PythonCHB, line 18: .split() already returns a list -- no need for the list()

Students/SSchwafel/session04/mailroom_dict.py
```diff
@@ -0,0 +1,196 @@
...
+            outfile = open('%s.txt'%i.replace(' ',''),'w' )
+
+            print i
+
+            outfile.write(send_email(i))
+    
+    elif user_input.upper() == 'D' or 'Q':
+
+        print 'Exiting...'
+        break
```
PythonCHB, line 196: this all looks good -- well done.

PR #49, opened 2014-10-26
Part of the assignment for Session04

imdavis:
I am leaving tomorrow morning on a trip for work.  I'm checking in all of this week's assignment except for the trigrams portion, which I haven't gotten to yet.  I'm hoping that I will find the time to work on that over the next several days while I am away on travel, but I'm not sure what kind of progress I will be able to make.


Students/imdavis/session04/mailroom/mailroomfunct.py
```diff
@@ -0,0 +1,138 @@
...
+
+    # write the message to a file also
+    filename = donorname+"-mail.txt"
+    try:
+        mailfile = open(filename, 'w')
+        mailfile.write(message)
+        mailfile.close()
+    except IOError:
+        print "Sorry...couldn't open %s"%filename
+
```
codedragon, line 127: Remember it is useful to just have one command in a try block, this way you can be fairly certain about what went wrong, if an exception is raised.

Students/imdavis/session04/mailroom/mailroomfunct.py
```diff
@@ -0,0 +1,138 @@
...
+def formattable(donor, total, ndonations, ave):
+    print"{:^20}{:^20}{:^20}{:^20}".format(donor, total, ndonations, ave)
+
```
codedragon, line 130: Nice table. :)

Students/imdavis/session04/paths.py
```diff
@@ -0,0 +1,30 @@
...
+print "\nMoving file:", filename
+print "  From:", srcdir
+print "  To  :", destdir
+
+f = open(srcpth+"/"+filename)
+fcpy = open(destpth+"/"+filename, 'w')
+for line in f:
+    fcpy.write(line)
+f.close()
+fcpy.close()
```
codedragon, line 30: Can also copy entire file at once. fcpy.write(f.read())

Students/imdavis/session04/trigram.py
```diff
@@ -0,0 +1,56 @@
...
+for filename in sys.argv[1:]:
+    for line in open(filename):
+        cleanline = line.lower().split()
+        if(len(cleanline) != 0):
+            for word in cleanline:
+                allwords.append(word.strip(strip))
+
+# build the trigram
+trigram = {}
+for i in range( len(allwords) - 3 ):
```
codedragon, line 24: You are skipping the last word. Remember if you do range(len(myvariable)) it includes the last word, so to get the end 3 words, you just need to subtract 2 to get the other 2: for i in range(len(allwords) - 2): 
Not sure how much sense that explanation makes, so probably best to play around with a list of a few numbers and see what you get when you loop through printing various groups of the numbers.

codedragon, line 24: Might want to think about adding back in some punctuation. :) But it is a very lovely run-on sentence, and a great start.

PR #51, opened 2014-10-27
Added session04 homework

meshmote:
Thanks!
Robert


Students/RPerkins/session04/test/testcopy.txt
```diff
@@ -0,0 +1 @@
+This is the test file content
```
PythonCHB, line 1: probably a good idea for you test file to be more than one line, just in case...

Students/RPerkins/session04/print_filepath.py
```diff
@@ -0,0 +1,7 @@
+__author__ = 'Robert W. Perkins'
+
+import pathlib
+
+pth = pathlib.Path('./')
```
PythonCHB, line 5: nice! pathlib is pretty cool, eh?

Students/RPerkins/session04/mailroom.py
```diff
@@ -0,0 +1,118 @@
+__author__ = 'Robert W. Perkins'
+
+
+def mk_dbase():
+    """Create data structure for donor list"""
+    # donor name = key: sum of donations, donation 1, donation 2, ...
```
PythonCHB, line 6: while it makes some sense to store a calculated value, it does mean you need to be careful to keep them in sync. You also might want to use a slightly more complex data structure, so that it's clear that the total is a differnt thing:

maybe:
 ndbase = {
        'Jeff McCarthy': (4000, [2500, 1000, 500]),

Students/RPerkins/session04/mailroom.py
```diff
@@ -0,0 +1,118 @@
...
+    except EOFError:
+        return None
+    except KeyboardInterrupt:
+        return None
+    return int(new_d)
+
+
+def print_email(p_name, p_donation):
+    """ Print thank you note for donation from p_name """
+    ltr_temp = {'Template1': 'Dear {name}, Thanks so much for your generous donation of ${donation}.  '
```
PythonCHB, line 43: why put this in a dict? are you anticipating being able to chose more than one template?

Students/RPerkins/session04/mailroom.py
```diff
@@ -0,0 +1,118 @@
...
+
+
+def safe_input():
+    try:
+        new_d = raw_input("Enter donation amount (must be numeric)-->")
+    except EOFError:
+        return None
+    except KeyboardInterrupt:
+        return None
+    return int(new_d)
```
PythonCHB, line 38: what will happen if the input is not an integer?

Students/RPerkins/session04/mailroom.py
```diff
@@ -0,0 +1,118 @@
...
+    ltr_temp = {'Template1': 'Dear {name}, Thanks so much for your generous donation of ${donation}.  '
+                             'It is greatly appreciated!'
+                }
+
+    print ltr_temp['Template1'].format(name=p_name, donation=p_donation)
+
+
+def app_record(app_name, app_dict):
+    """ Append an existing donor record """
+    app_donation = safe_input()
```
PythonCHB, line 52: what will happen if safe_input returns None?

Students/RPerkins/session04/mailroom.py
```diff
@@ -0,0 +1,118 @@
...
+
+    if name in donor_dict:
+        app_record(name, donor_dict)
+    else:
+        add_record(name, donor_dict)
+
+
+def write_efile(name, donation_list):
+    """write a donor email to disk named "name".txt"""
+    to_file = './%s.txt' % name
```
PythonCHB, line 85: the "./" is assumed anyway.

Students/RPerkins/session04/mailroom.py
```diff
@@ -0,0 +1,118 @@
...
+#def sum_element(key_dbase):
+    #"""set key for sorting on sum element of data structure"""
+    #return key_dbase[1]
+
+
+def mk_report(rep_dict):
+    """ Create a sorted list of donors"""
+    print 'Donor Name\t\t\tTotal Donations\t# of Donations\t\tAverage Donation'
+    #rep_dbase.sort(key=sum_element)
+    for j, k in rep_dict.items():
```
PythonCHB, line 100: how could you sort this?

Students/RPerkins/session04/mailroom.py
```diff
@@ -0,0 +1,118 @@
...
+        num_donations = len(k)-1
+        avg_donation = k[0]/(len(k)-1)
+        print '%s\t\t\t%s\t\t\t\t%s\t\t\t\t\t\t%s' % (j, k[0], num_donations, avg_donation)
```
PythonCHB, line 103: tabs are tricky -- different terminals, etc will expand them differently. 

better to use spaces -- and you can use the formatters to align it all for you.

PR #52, opened 2014-10-27
Session 04 homework and labs

gosunfish:


Students/CarolynEvans/session04/PathlibHomework.py
```diff
@@ -0,0 +1,26 @@
...
+
+    p = pathlib.Path('.')
+
+    for f in p.iterdir():
+        print f.absolute()
+
+
+def copyFile(filename):
+    outfile = 'copyof_' + filename
+    o = open(outfile, 'w')
```
PythonCHB, line 13: careful here -- this will open in text mode -- for a generaic file copier, you'd want to use binary mode: open(outfile, 'wb')
(also for reading).

Students/CarolynEvans/session04/labs.py
```diff
@@ -0,0 +1,89 @@
+
+def dict_lab():
+    # 1.
+    d = {'name': 'Chris', 'city': 'Seattle', 'cake': 'Chocolate'}
+    print d
+    d.pop('cake')
+    print d
+    d.setdefault('fruit', 'Mango')
+    print d
+    print d.viewkeys()
```
PythonCHB, line 10: did you catch the difference between viewkeys() and keys() ? In this case, either will work fine.

Students/CarolynEvans/session04/labs.py
```diff
@@ -0,0 +1,89 @@
...
+def dict_zip():
+    # 2.
+    l = list(range(0,16))
+    print l
+    print ''
+
+    m = []
+    for i in range(16):
+        m.append(hex(i))
+    print m
```
PythonCHB, line 24: nice -- can you now see how to do this with a list comprehension?

Students/CarolynEvans/session04/labs.py
```diff
@@ -0,0 +1,89 @@
...
+    s3 = set(range(0,21,3))
+    print s3
+    s4 = set(range(0,21,4))
+    print s4
+    print s3.issubset(s2)
+    print s4.issubset(s2)
+
+def set_lab2():
+    s = set(['P','y','t','h','o','n'])
+    s |= {'i'}
```
PythonCHB, line 50: interewting -- I hadn't thought to do it this way -- this gives you the union of the sets, then assigns that to the original. I"d probably jsut do:

s.add('i') 

here, but this certainly works!

Students/CarolynEvans/session04/labs.py
```diff
@@ -0,0 +1,89 @@
...
+    fs = set(['m','a','r','a','t','h','o','n'])
+    print fs
+    print 'Union: ', fs.union(s)
+    print 'Intersection: ', fs.intersection(s)
+
+def safe_input(label):
+    try:
+        safeinput = raw_input(label)
+        return safeinput
+    except:
```
PythonCHB, line 62: you really want to catch the particular Exceptions you are expecting here:

except (EOFError, KeyboardInterrupt) :

a bare except is rarely a good idea.

Students/CarolynEvans/session04/labs.py
```diff
@@ -0,0 +1,89 @@
...
+        # Strips leading spaces and converts to lower case for later grouping.
+        studentlanguages = student.split(":")[1].lower().strip()
+        languages.extend(studentlanguages.split(","))
+
+    languages.remove('languages')
+
+    # Jam packed stagement.
+    # Counts the number of occurences of each language string.
+    # Ignores keys (languages) that are empty strings or None.
+    languagedict = dict((language, languages.count(language)) for language in languages if language)
```
PythonCHB, line 80: very slick. nice! you could do this very slightly slicker iwtha  direct dict comprehension:

languagedict = { language: languages.count(language) for language in languages if language}

do note that languages.count is O( n ), so this is O( n^2 ) -- not a big deal if you know that you won't have to deal with huge data sets....

PythonCHB:
session04/ngram.txt is a really big file -- what's in it? Do you need to add it to the repo?

Students/CarolynEvans/session04/ngrams.py
```diff
@@ -0,0 +1,58 @@
...
+    This function builds a dictionary by looking at sets of words in the given file.
+    Words are are made up of the letters 'a' through 'z' and separated by a space.
+    Numbers and punctuation are removed.
+
+    Output is written to a file called ngram.txt.
+
+    :param nkey: Parameter nkey is the number of words to include in the dictionary key.
+    :param nvalue: Parameter nvalue is the number of words to include in the dictionary value.
+    :param filename: Parameter 'filename' is the name of the file containing the text to be processed.
+    """
```
PythonCHB, line 15: excellent docstring!

Students/CarolynEvans/session04/ngrams.py
```diff
@@ -0,0 +1,58 @@
...
+    import re # regex
+    from collections import defaultdict
+
+    wordlist = []
+    # Put each word into a list.
+    for line in open('sherlock.txt'):
+        # Use regex (re) to get rid of everything except alphabetic characters.
+        # Convert to lower case for grouping when you are playing around with the output file.
+        # Split into list by splitting on the space between each word.
+        wordlist.extend(re.sub("[^a-zA-Z ]+", "", line).lower().split(" "))
```
PythonCHB, line 26: at some point, you may want to do something with apostrophe's and hyphenated words. But nice job figuring out the regex module.

also, split() with no arguments splits on all whitespace, probably what you want in this case.

Students/CarolynEvans/session04/ngrams.py
```diff
@@ -0,0 +1,58 @@
...
+
+    # Get rid of empty entries
+    wordlist = filter(None, wordlist)
```
PythonCHB, line 29: very slick!

Students/CarolynEvans/session04/ngrams.py
```diff
@@ -0,0 +1,58 @@
...
+
+    # print wordlist
+
+    # Build the dictionary of ngrams.
+    ngramdict = defaultdict(list)
```
PythonCHB, line 34: nice discovery of defaultdict

Students/CarolynEvans/session04/ngrams.py
```diff
@@ -0,0 +1,58 @@
...
+    for i in range(0,len(wordlist) - nkey - nvalue):
+        # Build the key
+        k = str(wordlist[i])
```
PythonCHB, line 37: isn't wordlist[i] already a string?

Students/CarolynEvans/session04/ngrams.py
```diff
@@ -0,0 +1,58 @@
...
+        for j in range(i+1, i+nkey):
```
PythonCHB, line 38: you could do this with a slice -- something like:

k = ' '.join( wordlist[i:i+nkey] )

Students/CarolynEvans/session04/ngrams.py
```diff
@@ -0,0 +1,58 @@
+def ngrams(nkey=2,nvalue=1,filename='sherlock.txt'):
```
PythonCHB, line 1: nice job making it generalized to n keys and values ...

Students/CarolynEvans/session04/ngrams.py
```diff
@@ -0,0 +1,58 @@
...
+    ngramdict = defaultdict(list)
+    for i in range(0,len(wordlist) - nkey - nvalue):
+        # Build the key
+        k = str(wordlist[i])
+        for j in range(i+1, i+nkey):
+            k += ' ' + wordlist[j]
+
+        #Build the value
+        v = str(wordlist[i + nkey])
+        for j in range(i+1+nkey, i+nkey+nvalue):
```
PythonCHB, line 43: could use slice rather than loop here to.

Students/CarolynEvans/session04/ngrams.py
```diff
@@ -0,0 +1,58 @@
...
+    # print ' '
+    # print ngramdict
+
+    # Write the ngram dictionary to a file.
+    ngram = open('ngram.txt', 'w')
+
+    for item in ngramdict.iteritems():
+        ngram.write(str(item))
+
+ngrams(2,1)
```
PythonCHB, line 58: really nice job building up the ngramdict -- can you now use it to generte some "fake" text?

Students/CarolynEvans/session04/ngrams.py
```diff
@@ -0,0 +1,58 @@
...
+        ngramdict[k].append(v)
+
+    # print ' '
+    # print ngramdict
+
+    # Write the ngram dictionary to a file.
+    ngram = open('ngram.txt', 'w')
+
+    for item in ngramdict.iteritems():
+        ngram.write(str(item))
```
PythonCHB, line 56: note that you'e only written the items here -- if you want to do anything with this, you'd need the keys. Also, you might want to put newlines in there...

The file is too large to view on gitHub, but I think it's not quite what you would want.

Of course, for this exercise, you don't need to write ot to a file at all.

PythonCHB:
Please remove session04/ngram.txt:

git rm session04/ngram.txt

Then I can merge this -- I'd rather not merge that big ol' file into the main repo.

gosunfish:
ngram.txt is removed. 

----- Original Message -----

Students/CarolynEvans/session04/labs.py
gosunfish, line 10: No, I did not catch the difference between viewkeys() and keys(). Where can I get an explanation? 

----- Original Message -----

gosunfish, line 10: Nevermind. I found the answer in session 4. 

----- Original Message -----

gosunfish, line 10: When would one choose to create and use a view rather than use the underlying dictionary? 

----- Original Message -----

gosunfish:
Chris, 
For my final project, my boss requested I write a front end to a Redshift database we have on the AWS cloud. The database has 44 million rows in it, so the sorting and filtering will be done in the database rather than by the UI. All the UI has to do is accept the sorting and filtering parameters given by the user, send them to the database, and display the results. Does that sound like it would meet the class requirements? I don't think it would, but thought I would check with you first. I have to do this for work whether or not it meets the class project requirements, but I may need to do something else for the class project. 

Thanks, Carolyn Evans

PythonCHB:
Carolyn, 

I expect you'll find plenty to do in Python for this project. yes, the real work will be done in the DB, but you'll need to figure out how to call the db, manage the results, etc.

And useful is good.

PR #53, opened 2014-10-27
Added session04 homework

jerrytozhang:


Students/Hui Zhang/session04/dictsetlab.py
```diff
@@ -0,0 +1,86 @@
...
+
+
+"""
+   Part #2
+   Using the dict constructor and zip, build a dictionary of numbers from zero to fifteen and the hexadecimal
+   equivalent (string is fine).
+"""
+
+list1 = ['0', '1', '2', '3', '4', '5', '6', '7', '8', '9', '10', '11', '12', '13', '14', '15']
+list2 = ['0', '1', '2', '3', '4', '5', '6', '7', '8', '9', 'a', 'b', 'c', 'd', 'e', 'f']
```
PythonCHB, line 36: you can also get the hex value with hex(int)

but you got the zip idea just right.

Students/Hui Zhang/session04/dictsetlab.py
```diff
@@ -0,0 +1,86 @@
...
+
+
+
+
+"""
+   Part #3
+   Using the dictionary from item 1: Make a dictionary using the same keys but with the number of ‘t’s in each
+   value.
+"""
+dict1 = {'name': 'Chris', 'city': 'Seatle', 'cake': 'Chocolate'}
```
PythonCHB, line 48: note: "Seattle" has 2 ts ;-)

Students/Hui Zhang/session04/dictsetlab.py
```diff
@@ -0,0 +1,86 @@
...
+print dict1
+dict1['name'] = 'ttt'
```
PythonCHB, line 50: the idea here was to compute how many 't's there were in the values that were there.

Students/Hui Zhang/session04/dictsetlab.py
```diff
@@ -0,0 +1,86 @@
...
+
+
+"""
+    Part #4
+    Create sets s2, s3 and s4 that contain numbers from zero through twenty, divisible 2, 3 and 4.
+    Display the sets.
+    Display if s3 is a subset of s2 (False)
+    and if s4 is a subset of s2 (True).
+"""
+ s2 = set([2, 3, 4, 8, 9, 12, 15, 18])
```
PythonCHB, line 65: all correct, but how would you compute these sets, rather than hard-coding them?

Students/Hui Zhang/session04/exceptionslab.py
```diff
@@ -0,0 +1,12 @@
+"""
+   Create a wrapper function, perhaps safe_input() that returns None rather rather than raising
+   these exceptions, when the user enters ^C for Keyboard Interrupt, or ^D (^Z on Windows) for
+   End Of File.
+"""
+def safe-input():
+    try:
+        test_exception = raw_input("Please type something and press Ctrl+C\n Or\n Just press Ctrl+D without typing anything\n to test exceptions: ")
```
PythonCHB, line 8: you got the exception handling -- but this would be more useful if you could pass the message to the user in to this function, so it could be used in place of raw_input()

Students/Hui Zhang/session04/fileslab.py
```diff
@@ -0,0 +1,63 @@
...
+language_list_1 = []
+language_list_stripped = []
+#
+# open students.txt
+file = open('students.txt')
+# remove students name from string_name_language list, and add all languages to a new list called language_list.
+string_name_language = file.readlines()
+for item in string_name_language:
+    language_start_position = item.index(':')
+    language_items = item[language_start_position + 1: -1]
```
PythonCHB, line 15: you can just leave the bare colon to get "to the end of the sequence":

language_items = item[language_start_position + 1:]

but take a look at other string methods tha make this a bit easier -- string.split() is pretty handy here.

Students/Hui Zhang/session04/fileslab.py
```diff
@@ -0,0 +1,63 @@
...
+# remove first element, strip whitespace from elements and remove empty elements from language_list list.
+# make a list called language_list_stripped, in which each element contains one or multiple languages.
+for item1 in language_list:
+    item2 = item1.strip()
+    if item2 != 'languages' and item2 != '':
+        language_list_stripped.append(item2)
+
+# make a new list called language_list_1, in which, each element is a list, and each element in the inner list contains ONE language.
+for item3 in language_list_stripped:
+    item4 = item3.split()
```
PythonCHB, line 27: you can pass in a string to split on, if not whitespace -- i.e. 

item3.split(',')

Students/Hui Zhang/session04/fileslab.py
```diff
@@ -0,0 +1,63 @@
...
+    language_list_1.append(item4)
+
+# convert language_list_1 to a new list called language_list_2, in which, each element is a language.
+item7 = ''
+for item5 in language_list_1:
+    # convert a list to a string
+    item6 = ''.join(item5)
```
PythonCHB, line 34: why split in the previous step, just to re-join here?

Students/Hui Zhang/session04/fileslab.py
```diff
@@ -0,0 +1,63 @@
...
+# remove all empty elements from the list: language_list_2.
+while True:
+    try:
+        empty_element_positon = language_list_2.index('')
+        language_list_2.pop(empty_element_positon)
+    except ValueError:
+        break
+
+language_list_3 = list(set(language_list_2))
+
```
PythonCHB, line 52: why not just keep it in the set? or, for that matter, put them in a set in teh first place as you go through the lines in the file.

Students/Hui Zhang/session04/fileslab.py
```diff
@@ -0,0 +1,63 @@
...
+print 'List all languages that students know :\n'
+for item9 in language_list_3:
+    print item9
+
+# get the number of students on each language listed in language_list_3.
+counter = collections.Counter(language_list_3)
```
PythonCHB, line 58: nice find! very slick!

but did this work? it looks like language_list_3 was passed through set() and thus is unique, so the count will always be one. 

Uf you're going to use Counter, which is a great idea, then you can jsut use the Counter object, and not do the set at all.

Students/Hui Zhang/session04/pathandfile.py
```diff
@@ -0,0 +1,6 @@
+
+import pathlib
+pth = pathlib.Path('./')
+path_string = str(pth.absolute()) + '/'
+for f in pth.iterdir():
+        print path_string + str(f)
```
PythonCHB, line 6: nice use of pathlib

PythonCHB:
You're getting it. 

But for the filelab -- if it seems like you are doing too much work, you probably are. Python has powerful string processing, so you should be able to do all that in fewer, easier steps in one loop.

take a look at my example. Though your use of collections.Counter() is nice -- I hadn't thought of that.

PR #55, opened 2014-10-28
Session 4 Homework

salimhamed:
Here's my homework for session 04.  Please review at your leisure.


salimhamed:
Also added session 5 homework.

Students/salim/session04/kata_14.py
```diff
@@ -0,0 +1,143 @@
...
+    return a_list
+
+
+def build_trigram(all_words):
+    """Return a trigram dictionary."""
+    trigram_dict = dict()
+    for idx, word in enumerate(all_words):
+
+        # only add words to dict if there are three words left
+        if idx < len(all_words) - 3:
```
codedragon, line 44: Since you are using a less than sign, this will actually stop when there are four words left, so to get all trigrams possible you would use 'if idx < len(all_words) - 2:'

Students/salim/session04/kata_14.py
```diff
@@ -0,0 +1,143 @@
...
+
+    # get trigram list
+    trigram_list = get_trigram_output(trigram_dict,
+                                      start_words,
+                                      num_of_words_to_print)
+
+    # print trigram list
+    print_string = ('{:s} ' * len(trigram_list))[:-1]
+    print 'Trigram of {:d} words:'.format(len(trigram_list))
+    print print_string.format(*trigram_list)
```
codedragon, line 143: Nice Solution. For this particular text, probably want to look for the line that starts 'End of the Project Gutenberg EBook' and stop collecting lines at that point.

Students/salim/session04/mailroom2.py
```diff
@@ -0,0 +1,161 @@
...
+        # add donation to history
+        if donor_in_list:
+            send_mail(response, amount)
+            add_existing_donor_to_list(response, amount, donor_list)
+            return
+        else:
+            send_mail(response, amount)
+            add_new_donor_to_list(response, amount, donor_list)
+            return
+
```
codedragon, line 80: Notice you have 2 of the exact same calls in your if and else statements. These can be done outside of your conditional so you are only calling them once.

codedragon, line 80: This is a good solution, but you should consider how a dictionary might be used instead. And, you did. :)

Students/salim/session05/test_mailroom.py
```diff
@@ -0,0 +1,26 @@
+#!/usr/local/bin/python
+
+
+import mailroom3 as mail
+
+
+def test_create_donation_list():
+    assert type(mail.create_donation_list()) == dict
+
```
codedragon, line 9: Great start on testing. Should probably test actually sticking a name and donation into the dictionary as well.

PR #58, opened 2014-10-29
adding trigram homework for alexg

ghost:


Students/alexg/session04/trigram.py
```diff
@@ -0,0 +1,34 @@
...
+#converts text to dictionary
+for x in range(len(text)):
+        try:
+            if not text[x]+' '+text[x+1] in collection:
+                collection[text[x]+' '+text[x+1]]=text[x+2].split()
+            else:
+                collection[text[x]+' '+text[x+1]]+=text[x+2].split()
+        except IndexError:
+            print 'done'
+
```
codedragon, line 21: Notice that you are adding and removing spaces quite a bit in this code. One way to deal with this would be to add your two words as your dictionary key as a tuple, rather than a string. To avoid the IndexError, use 'for x in range(len(text) - 2)'.

Students/alexg/session04/trigram.py
```diff
@@ -0,0 +1,34 @@
...
+temp2 = temp.split()
+x=0
+
+#iterates through dictionary and creates new text
+while x!=1:
+        try:
+            temp2+=random.choice(collection[temp2[-2]+' '+temp2[-1]]).split()
+        except KeyError:
+            print ' '.join(temp2)
+            x=1
```
codedragon, line 34: This is a nice, concise answer, well done. Your output does end up with quite random and often nonsensical punctuation. Assuming you were trying to make this somewhat realistic, you would probably want to remove the punctuation, and then add some back in.

PR #60, opened 2014-10-29
Initial commit

ARSimmons:
testing if this works :)


Students/ASimmons/Session02/test_ack.py
```diff
@@ -0,0 +1,55 @@
+__author__ = 'Ari'
+
+from ack import ackerman
+import unittest
+
+# Note to Chris: So, I was experimenting with classes
+# and I've had this problem in the past where I try to run
```
PythonCHB, line 7: getting ahead of yourself!

PR #61, opened 2014-11-01
Session02

pythonandme:
just testing


Students/Danielle_Marcos/session01/break_me.py
```diff
@@ -0,0 +1,17 @@
+def myNameError():
+	a > 0
+
+def myTypeError():
+	print 1 + 'ten'
+
+def mySyntaxError():
```
PythonCHB, line 7: looks good!

Students/Danielle_Marcos/session03/Dictionary_Lab.py
```diff
@@ -0,0 +1,23 @@
...
+mydict.pop('cake')
+print mydict
+
+mydict['fruit'] = 'Mango'
+print mydict
+
+print mydict.keys()
+print mydict.values()
+
+if 'cake' in mydict.keys():
```
PythonCHB, line 13: this works, of course, but:

 'cake' in mydict

will work as well -- the keys are assumed.

and no need for the if -- key in dict will return True or False anyway.

Students/Danielle_Marcos/session03/Dictionary_Lab.py
```diff
@@ -0,0 +1,23 @@
...
+	print True
+else: 
+	print False
+
+if 'Mango' in mydict.values():
+	print True
+else:
+	print False
+
+mydict2 = {"0"}
```
PythonCHB, line 23: where are you going with this?

PR #63, opened 2014-11-02
Session 4/5/6 Combined Pull

Anastomose:
Accidentally wiped Session 4 branch before it could be merged on remote, but changes have been rolled forward into Session 5 branch.  New updates include Session 5 labs for testing, comprehension, and argument passing.


Anastomose:
Fixed the merge conflict!  Merge me in!

PythonCHB:
Lots here (moslty because we've gotten behind on reviewing...) -- is there anything in particular you want reviewed, to make sure we don't miss it?

Students/ebuer/session_06/render/test_html_outputXX.html
```diff
@@ -0,0 +1,4 @@
+<html>
```
PythonCHB, line 1: maybe you don't need this added to the repo?

Students/ebuer/session_06/render/html_render.py
```diff
@@ -0,0 +1,39 @@
+"""
+html_render.py
+
+Questions: how do I inspect class and object attributes to see if the correct value has been ascribed?
```
PythonCHB, line 4: generally not -- trust your user to use it rigth -- if not, then they'll get an error at run-time.

Students/ebuer/session_06/render/html_render.py
```diff
@@ -0,0 +1,39 @@
...
+
+"""
+
+# starting render classes
+
+
+class Element(object):
+    indent = 0  # indent level for tag
+    tag = [u"<html>", u"</html>"]
```
PythonCHB, line 13: you sure you want to do unicode?? It's a good idea, but we haven't covered it yet.

PythonCHB, line 13: I like storing them in a pair like this -- though you may find you need to render the start tag differently later on....

Students/ebuer/session_06/render/html_render.py
```diff
@@ -0,0 +1,39 @@
...
+
+    def __init__(self, content=None):
+        self.content = content
+
+    def append(self, new_content=""):
+        if self.content is None:
+            self.content = [new_content] # create a list to render ea. element
+        else:
+            self.content.append(new_content)
```
PythonCHB, line 22: I think this is going to break is the user passes in a string to teh **init**, then wants to append more.

your **init** should always create a list, so you know it's there.

Students/ebuer/session_06/render/html_render.py
```diff
@@ -0,0 +1,39 @@
...
+
+    # renders the tag and the strings in the content
+    def render(self, file_out, ind=""):
+        self.ind = ind
+        self.indent = 1  # subclass to override class indent (?)
```
PythonCHB, line 27: yeah, that's the idea -- for the most part, all tags will use the same indent. If one needs somethign different, you can override in that subclass.

Students/ebuer/session_06/render/html_render.py
```diff
@@ -0,0 +1,39 @@
...
+
+        # content will need to be read in one line at a time and rendered
+
+        self.file_out = file_out.write(u'{indent}{tag}\n'
+            .format(indent=Element.indent*self.ind, tag=Element.tag[0]))
```
PythonCHB, line 32: you want to use self.indent here... it will find the class attribute.

Students/ebuer/session_06/render/html_render.py
```diff
@@ -0,0 +1,39 @@
...
+
+    def append(self, new_content=""):
+        if self.content is None:
+            self.content = [new_content] # create a list to render ea. element
+        else:
+            self.content.append(new_content)
+
+    # renders the tag and the strings in the content
+    def render(self, file_out, ind=""):
+        self.ind = ind
```
PythonCHB, line 26: any need to assign to self?

Students/ebuer/session_06/render/html_render.py
PythonCHB, line 32: this implies that self.ind is a number (level of indentation) -- good way to go, but the default is set to an emptly string -- should that be zero for the default?

Students/ebuer/session_06/lambdalab.py
```diff
@@ -0,0 +1,8 @@
+"""
+lambdalab.py
+"""
+
+
+def function_builder(numrng):
+    list = [lambda n, i=n: n ** 2 + i for n in range(numrng)]
```
PythonCHB, line 7: you are using "n" in two contexts here -- It should work, but it's confusing -- probably worth a different name.

Students/ebuer/session_06/funcfile/funcfile.py
```diff
@@ -0,0 +1,54 @@
...
+# Read in the original file and return contents
+def gettext(infile):  # ='junkfile.txt'
+    fid = StringIO()
+    with open(infile, 'r', encoding='UTF-8') as f:
+        fid = f.readlines()
+        return fid
+
+# Apply map lambda function to clean the text
+def spstrip(fid):
+    return map(lambda x: x.strip(' \n'), fid)
```
PythonCHB, line 28: don't need the lambda -- str.strip is already a function:

map(str.strip, fid)

also note that a file object is an iterator, so you can just pass the file object in to the map(), without the readlines() call.

Students/ebuer/session_06/funcfile/funcfile.py
```diff
@@ -0,0 +1,54 @@
...
+
+# use a list comprehension instead of the map function
+def compstrip(fid):
+    return [x.strip(' \n') for x in fid]
```
PythonCHB, line 32: yup!

Students/ebuer/session_06/funcfile/funcfile.py
```diff
@@ -0,0 +1,54 @@
...
+
+# print a new file
+def pttext(outfile, intext):
+    with open(outfile, 'w', encoding='UTF-8') as f:
```
PythonCHB, line 36: nice use of with, by the way....

Students/ebuer/session_06/funcfile/funcfile.py
```diff
@@ -0,0 +1,54 @@
...
+        [f.write(u'{line}\n'.format(line=l)) for l in intext]
+
```
PythonCHB, line 38: slick!

Students/ebuer/session_05/test_lab.py
```diff
@@ -0,0 +1,55 @@
...
+
+# adding test data for py.test module
+
+test_data = [((3, 5), False),
+             ((7, 9), True),
+             ((10, 5), True),
+             ((20, 1), True)]
+
+# borrowing Chris's use of the parametrize attribute
+@pytest.mark.parametrize(("input", "result"), test_data)
```
PythonCHB, line 31: nicely done!

PythonCHB:
Looking good -- let me know if you want anything else reviewed -- a lot here, so I looked at the newer stuff.

PR #64, opened 2014-11-03
Homework for session 5 plus session 3 and 4 stuff

agrootsky:


Students/apklock/session05/list_comp.py
```diff
@@ -0,0 +1,90 @@
...
+
+
+# Moving along
+
+>>> comprehension = { x for x in 'aabbbcccc'}
+
+# What is the output of:
+
+>>> comprehension
+'a', 'b', 'c' # I think...
```
PythonCHB, line 68: yup!

Students/apklock/session05/list_comp.py
```diff
@@ -0,0 +1,90 @@
...
+>>> comprehension = [ skit * number for number, skit in list_of_tuples ]
+
+# What is the output of:
+
+>>> comprehension[0]
+'lumberjack'
+
+>>> len(comprehension[2])
+11 # I'm not sure about this one, length of 'inquisition'?
+
```
PythonCHB, line 42: nope -- index 2 is 'spam', the the comp multiplies it times 4...

Students/apklock/session05/dict_set_comp.py
```diff
@@ -0,0 +1,76 @@
...
+
+# creating a sequence of all 3 sets
+
+s2_new = set()
+s3_new = set()
+s4_new = set()
+
+all_s_set = [[s2_new], [s3_new], [s4_new]]
+
+for i in range(21):
```
PythonCHB, line 68: could you do this without repeating an code? i.e the smae code for i%2, i%3, etc...

PythonCHB:
It looks like mostly my solutions for mailroom and trigrams .. do make sure you understand these and could do something like that yourself from scratch.

PR #68, opened 2014-11-04
Added Session04 and Session05 folders

QuietMike:
Hello,

In session05 in my folder, I posted a program for the trigram homework. I created it without looking at the solution you posted (and forgot most of the details of what we reviewed in the previous class). Please, let me know what you think if you have time to look at it.
Thank you very much!
Michel.

### DIFF
--- a/Students/michel/Session05/trigrams.py
+++ b/Students/michel/Session05/trigrams.py
@@ -1,0 +1,131 @@
+# -*- coding: utf-8 -*-
+"""
+Created on Sun Nov 02 16:48:08 2014
+
+@author: Michel
+"""
+
+def readText(fileName):
+    ''' 
+    Reads the raw text from a text file 
+    fileName is a string with path and file name
+    Returns a string with the text read in it
+    '''
+    sourceText = open(fileName, 'r')
+    text = sourceText.read()
+    sourceText.close()
+    return text
+
+    
+def cleanText(text):
+    '''
+    Cleans the text by transforming it into lower cases
+    removing all punctuation except period
+    Returns a clean text ready for processing
+    '''
+    text = string.lower(text)
+    text = string.replace(text, '\n', ' ')
+    text = string.replace(text, '--', ' ')
+    charList = '''!"#$%&()*+,-/:;<=>?@[\]^_`{|}~0123456789'''
+    for i in charList:
+        text = string.replace(text, i, ' ')
+    text = ' '.join(text.split())
+    return text
+    
+
+def buildDict(text):
+    '''
+    Splits the text in sentences and builts pairs of words for each sentence
+    Returns a dictionary of trigrams with all pairs of words and their 
+    list of following words
+    '''
+    sentenceList = []
+    pairsDict = {}
+    wordCount = 0
+    sentenceList = text.split('.')
+    
+    for sentence in sentenceList:
+        wordList = []
+        sentence = string.strip(sentence, ' ')
+        sentence = string.replace(sentence, '.', '')
+        wordList = string.split(sentence, ' ')
+        wordCount = len(wordList)        
+        
+        for i in range(wordCount):
+            if  (wordCount < 2) or (i > wordCount - 3):
+                break
+            else:
+                wordTuple = ()                          
+                wordTuple = (wordList[i], wordList[i+1])
+                pairsDict.setdefault(wordTuple, []) 
+                if wordList[i+2] not in pairsDict[wordTuple]:
+                    pairsDict[wordTuple].append(wordList[i+2])
+                    
+    return pairsDict
+
+
+def createSentence(pairsDict):
+    '''
+    Creates one trigram sentence
+    If a pair of words is not found in the dictionary, the sentence ends
+    Returns a string sentence    
+    '''
+    sentenceLength = random.randint(5, 15)
+    startKey = ()
+    sentence = []
+    textSentence = ''
+    
+    keyList = pairsDict.keys()  
+    
+    # Starts a sentence
+    startKey = keyList[random.randint(0, len(pairsDict)-1)]
+    sentence = [startKey[n] for n in range(2)]
+    word = pairsDict[startKey][random.randint(0, len(pairsDict[startKey])-1)]
+    if word == 'i':
+        word = 'I'
+    sentence.append(word)
+    ending =(':', '!', '?', '.', '...')
+    
+    # Continues a sentence
+    for j in range(sentenceLength-2):
+        tempKey = (sentence[-2], sentence[-1])
+        if pairsDict.has_key(tempKey):
+            word = pairsDict[tempKey][random.randint(0, len(pairsDict[tempKey])-1)]
+            sentence.append(word)
+        else:
+            break
+        
+    textSentence = ' '.join(sentence)
+    textSentence = string.capitalize(textSentence) + \
+    ending[random.randint(0, len(ending)-1)]
+
+    return textSentence
+
+        
+def createText(pairsDict):
+    '''
+    Creates a random text based on the dictionary of trigrams
+    Text is made of 40 to 50 random sentences of variable length.
+    Returns a string text
+    '''
+    textLength = random.randint(40, 50)
+    text = ''
+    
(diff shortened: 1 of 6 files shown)


Students/michel/Session05/trigrams.py
```diff
@@ -0,0 +1,131 @@
+# -*- coding: utf-8 -*-
+"""
+Created on Sun Nov 02 16:48:08 2014
+
+@author: Michel
+"""
+
+def readText(fileName):
```
PythonCHB, line 8: not sure it's worth a function for this -- it could just be:

text = open(fileName).read()

Students/michel/Session05/trigrams.py
```diff
@@ -0,0 +1,131 @@
...
+    Cleans the text by transforming it into lower cases
+    removing all punctuation except period
+    Returns a clean text ready for processing
+    '''
+    text = string.lower(text)
+    text = string.replace(text, '\n', ' ')
+    text = string.replace(text, '--', ' ')
+    charList = '''!"#$%&()*+,-/:;<=>?@[\]^_`{|}~0123456789'''
+    for i in charList:
+        text = string.replace(text, i, ' ')
```
PythonCHB, line 31: looks good, but it is a bit slow once you get to a lot of characters to replace -- that's why the .translate() function is there. But this is great for now -- you can optimize it some day if you ever need to.

Students/michel/Session05/trigrams.py
```diff
@@ -0,0 +1,131 @@
...
+def cleanText(text):
+    '''
+    Cleans the text by transforming it into lower cases
+    removing all punctuation except period
+    Returns a clean text ready for processing
+    '''
+    text = string.lower(text)
+    text = string.replace(text, '\n', ' ')
+    text = string.replace(text, '--', ' ')
+    charList = '''!"#$%&()*+,-/:;<=>?@[\]^_`{|}~0123456789'''
```
PythonCHB, line 29: do  you want to get rid of all numbers? I'm not sure what's in this text but maybe not...

Students/michel/Session05/trigrams.py
```diff
@@ -0,0 +1,131 @@
...
+    '''
+    sentenceList = []
+    pairsDict = {}
+    wordCount = 0
+    sentenceList = text.split('.')
+    
+    for sentence in sentenceList:
+        wordList = []
+        sentence = string.strip(sentence, ' ')
+        sentence = string.replace(sentence, '.', '')
```
PythonCHB, line 50: are there any periods left? the split('.') should have removed them.

Students/michel/Session05/trigrams.py
```diff
@@ -0,0 +1,131 @@
...
+        wordList = string.split(sentence, ' ')
+        wordCount = len(wordList)        
+        
+        for i in range(wordCount):
+            if  (wordCount < 2) or (i > wordCount - 3):
```
PythonCHB, line 55: why not just range(wordCount-2) ?

Students/michel/Session05/trigrams.py
```diff
@@ -0,0 +1,131 @@
...
+                break
+            else:
+                wordTuple = ()
```
PythonCHB, line 58: this empty one just gets thrown away when you make the new one. Remember that names are bound to objects in python -- so when, in the next line, you do:
wordTuple = ....

the name wordTuple gets re-bound to a new object -- so having it set to an empty tuple does absolutely nothing....

Students/michel/Session05/trigrams.py
```diff
@@ -0,0 +1,131 @@
...
+        
+        for i in range(wordCount):
+            if  (wordCount < 2) or (i > wordCount - 3):
+                break
+            else:
+                wordTuple = ()                          
+                wordTuple = (wordList[i], wordList[i+1])
+                pairsDict.setdefault(wordTuple, []) 
+                if wordList[i+2] not in pairsDict[wordTuple]:
+                    pairsDict[wordTuple].append(wordList[i+2])
```
PythonCHB, line 62: if you don't want to double-count, then maybe use a set, rather than a list? (I just put them all in, so that works that followed more frequently, were also more likely to be selected later..

Students/michel/Session05/trigrams.py
```diff
@@ -0,0 +1,131 @@
...
+
+
+def createSentence(pairsDict):
+    '''
+    Creates one trigram sentence
+    If a pair of words is not found in the dictionary, the sentence ends
+    Returns a string sentence    
+    '''
+    sentenceLength = random.randint(5, 15)
+    startKey = ()
```
PythonCHB, line 74: again, no need to initialize to an empty tuple...

Students/michel/Session05/trigrams.py
```diff
@@ -0,0 +1,131 @@
...
+    sentence = []
+    textSentence = ''
```
PythonCHB, line 76: or an empty string...

Students/michel/Session05/trigrams.py
```diff
@@ -0,0 +1,131 @@
...
+    
+    keyList = pairsDict.keys()  
+    
+    # Starts a sentence
+    startKey = keyList[random.randint(0, len(pairsDict)-1)]
+    sentence = [startKey[n] for n in range(2)]
```
PythonCHB, line 82: this makes a new list, too -- so no need to have initialized an empty list.

Students/michel/Session05/trigrams.py
```diff
@@ -0,0 +1,131 @@
...
+    word = pairsDict[startKey][random.randint(0, len(pairsDict[startKey])-1)]
+    if word == 'i':
+        word = 'I'
```
PythonCHB, line 85: nice touch!

Students/michel/Session05/trigrams.py
```diff
@@ -0,0 +1,131 @@
...
+        tempKey = (sentence[-2], sentence[-1])
+        if pairsDict.has_key(tempKey):
+            word = pairsDict[tempKey][random.randint(0, len(pairsDict[tempKey])-1)]
+            sentence.append(word)
+        else:
+            break
+        
+    textSentence = ' '.join(sentence)
+    textSentence = string.capitalize(textSentence) + \
+    ending[random.randint(0, len(ending)-1)]
```
PythonCHB, line 100: another nice touch!

Students/michel/Session05/trigrams.py
```diff
@@ -0,0 +1,131 @@
...
+    Creates a random text based on the dictionary of trigrams
+    Text is made of 40 to 50 random sentences of variable length.
+    Returns a string text
+    '''
+    textLength = random.randint(40, 50)
+    text = ''
+    
+    # Creates a text made of sentences
+    for i in range(textLength):
+        textSentence = ''
```
PythonCHB, line 116: one more unnecessary initialization...

PythonCHB:
Nice job on trigrams -- just need to work on that initializing habit!

QuietMike:
Hello,
Thank you very much for reviewing my code and giving me advices on how to improve it! 

So, yes, I like initializations because they allow me to list my variables at the beginning. I got that habit from Pascal and Matlab... I'll try to work on it. Python sometimes requires it though.

 The reason why I took the numbers out is because of the chapters and sections of the book which are numbered. I also left the periods in the text because I tried to build pairs that were within a sentence as an improvement you suggested during the review. I am not sure I was successful though.

Michel.

PythonCHB:
> I got that habit from Pascal and Matlab

yup, some languages do require it (though not Matlab, as it happens)

> Python sometimes requires it though.

ah -- that's the key. it's not "pythonic", but more importantly, this class is the time to make sure that you really understand how name binding and mutable vs. immutable objects work. And that's the issue here:

you never need to initialize a name that you are going to re-bind, like::

  a = 0
  a = x + 5

 -- that a = 0 is completely useless

but you sometimes need to create an empty object that you are then going to mutate (change in place):

  a_list = []
  a_list.append(something)

note that you are not using the "=" in the second line -- you are not re-binding that name to something else.

PR #70, opened 2014-11-05
Session02

pythonandme:
This is the actual file I should have submitted when I was cleaning my github folders and files on Saturday during office hours. Ooops, sorry :)


Students/Danielle_Marcos/session04/Dictionary_Lab.py
```diff
@@ -0,0 +1,73 @@
...
+mydict2 = dict(myzip)
+print mydict2
+
+mydict3 = mydict.keys()
+print mydict3
+
+
+for i in mydict.values():
+	print i
+	if 't' in i:
```
PythonCHB, line 44: if you are going to use count() -- no need to check for 't' first..

Students/Danielle_Marcos/session04/Dictionary_Lab.py
```diff
@@ -0,0 +1,73 @@
...
+		print i
+		print i.count('t')
+
+s2 = set([2,4,6,8,10,12,14,16,18,20])
```
PythonCHB, line 48: how would you compute these, rather than hand-writing them in?

Students/Danielle_Marcos/session04/Dictionary_Lab.py
```diff
@@ -0,0 +1,73 @@
...
+s4 = set([4,8,12,16,20])
+
+print s2
+print s3
+print s4
+
+print s3.issubset(s2) 
+print s4.issubset(s2)
+
+s5 = set(['P','y','t','h','o','n'])
```
PythonCHB, line 59: a string is a sequence of characters, so this can just be:
s5 = set('Python')

PR #72, opened 2014-11-09
Homework 6 submitted

voite1:
Almost finished homework 6.  Only part 8 of the last assignment left...


Students/A.Kramer/session06/functional_files.py
```diff
@@ -0,0 +1,57 @@
...
+        print "Using default file 'text.txt' located in the same directory with '%s' script" % os.path.basename(sys.argv[0])
+    
+    # Prompt for a file name (if any)
+    new_file_name = ""
+    answer = raw_input("Would you like to create a new file [Y/N]: ")
+    if answer.lower() == "y":
+        new_file_name = raw_input("Enter new file name (no spaces): ")
+        if len(new_file_name.split()) > 1:
+            print "You have entered a filename containing space. Buy. Try better next time. \nBye."
+            # force to restart if space is detected in the filename
```
PythonCHB, line 20: spaces are legal in filenames -- though a agree -- bad idea!

Students/A.Kramer/session06/functional_files.py
```diff
@@ -0,0 +1,57 @@
...
+            return
+    
+    # Create temporary file
+    temp_file = os.path.join(os.getcwd(), "temp.temp.txt")
+    
+    # Open files for reading/writing    
+    try:
+        with open(path_to_a_file, "rb") as f:
```
PythonCHB, line 28: nice job finding "with" -- we'll get there in class...

Students/A.Kramer/session06/functional_files.py
```diff
@@ -0,0 +1,57 @@
...
+            lst = f.read().splitlines()
+            
+        # strip white space
+        g = map(lambda x: x.strip(), lst) # g = [x.strip() for x in lst] # comprehension
```
PythonCHB, line 32: you don't actually need a lambda here -- str.strip is already a function:

g = map(str.strip, lst)

would do.

Students/A.Kramer/session06/functional_files.py
```diff
@@ -0,0 +1,57 @@
...
+        
+        # write stripped lines into the temp.temp.txt file
+        with open(temp_file, "wb") as nf:
+            for item in g:
+                if item:
```
PythonCHB, line 37: this will remove empty lines -- do you want that?

Students/A.Kramer/session06/functional_files.py
```diff
@@ -0,0 +1,57 @@
...
+                    nf.write(item)
```
PythonCHB, line 38: I think the strip() above will have removed the newlines, so you may get one long line from this...

Students/A.Kramer/session06/functional_files.py
```diff
@@ -0,0 +1,57 @@
...
+                   
+    # handle IOErrors                
+    except IOError as e:
+        print "IOError, debugging required", e
+        return
+
+    # if new name is given, rename temp file to the new name
+    # if no new name, override the existing file
+    if new_file_name:
```
PythonCHB, line 47: why not just write to the new file in the first place?

Students/A.Kramer/session06/html_render/html_render.py
```diff
@@ -0,0 +1,184 @@
...
+        file_out.write(self.starttag)
+        for line in self.lst:
+            try:
+                line.render(file_out)
+            except AttributeError as e:
+                file_out.write(self.ind + ind + line + "\n")
+        file_out.write(self.endtag)
+
+class Html(Element):
+    def __init__(self, content=None, id=None, style=None):
```
PythonCHB, line 32: not a lot of code, but still not good to repeat yourself wuite so much -- if all instances of a class will have the same values, make them class attributes: i.e tag, indent, etc.

Students/A.Kramer/session06/html_render/html_render.py
```diff
@@ -0,0 +1,184 @@
...
+
+class Body(Element):
+    def __init__(self, content=None, id=None, style=None):
+        Element.__init__(self, content)
+        self.tag = "<body>"
+        self.starttag = self.tag + "\n"
+        self.endtag = self.tag[:1] + "/" + self.tag[1:] + "\n"
+        
+class P(Element):
+    def __init__(self, content=None, id=None, style=None):
```
PythonCHB, line 46: the idea here was to be abel to take ANY attributes, not just these particular ones -- they were just examples.

Others did this too -- sorry the instructions weren't more clear.

Students/A.Kramer/session06/html_render/html_render.py
```diff
@@ -0,0 +1,184 @@
...
+
+    def render(self, file_out, ind=""):
+        file_out.write(self.tag)
+
+    def append(self, new_content):
+        raise NotImplementedError
+
+class Hr(SelfClosingTag):
+    def __init__(self, content=None, id=None, style=None):
+        SelfClosingTag.__init__(self, content)
```
PythonCHB, line 98: id and style have been lost here.

Students/A.Kramer/session06/html_render/html_render.py
```diff
@@ -0,0 +1,184 @@
...
+            self.lst.append(content)
+            
+    def append(self, new_content):
+        self.lst.append(new_content)
+
+    def render(self, file_out, ind=""):
+        file_out.write(self.starttag)
+        for line in self.lst:
+            try:
+                line.render(file_out)
```
PythonCHB, line 26: does the nested indentation work outright here? I think you need to pass the current indentation level into the line render method.

Students/A.Kramer/session06/html_render/html_render.py
```diff
@@ -0,0 +1,184 @@
...
+class P(Element):
+    def __init__(self, content=None, id=None, style=None):
+        Element.__init__(self, content, id, style)
+        self.tag = "<P>"
+        self.id = ""
+        self.style = ""
+        if id is not None:
+            self.id = " id=\"" + id + "\"" 
+        if style is not None:
+            self.style = " style=\"" + style + "\""
```
PythonCHB, line 54: python can use either ' or " for quoting strings, so you can do:

 self.style = ' style='"' + style + '"'

rather than escape the quotes -- kind of nifty.

PR #74, opened 2014-11-18
This includes Week 6 homework

saschwafel:
I renamed the html_render.py file from the one I was noting down during optional class, so that's the reason for all the deletions and insertions. I tried to create it without looking at the notes as much as possible, but classes are pretty confusing, so a lot of it is very similar.

### DIFF
--- a/Students/SSchwafel/session06/functional_files.py
+++ b/Students/SSchwafel/session06/functional_files.py
@@ -1,0 +1,59 @@
+#!/usr/bin/python
+
+import os
+import sys
+
+filename = sys.argv[1:]
+
+def renamer(file_to_be_renamed):
+
+    path = os.getcwd()
+
+    files = os.listdir(path)
+
+    if file_to_be_renamed in files:
+
+        new_name = raw_input( "The current filename is {}, what would you like to call it instead? ".format(file_to_be_renamed))
+
+        os.rename(file_to_be_renamed, new_name)
+    
+    else:
+        print "It doesn't look like that file is present"
+
+map(renamer,filename)
+
+
+###UNUSED CODE FROM BEFORE I READ THE ASSIGNMENT ALL THE WAY###
+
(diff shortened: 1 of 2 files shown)


Students/SSchwafel/session06/functional_files.py
```diff
@@ -0,0 +1,59 @@
...
+
+filename = sys.argv[1:]
+
+def renamer(file_to_be_renamed):
+
+    path = os.getcwd()
+
+    files = os.listdir(path)
+
+    if file_to_be_renamed in files:
```
PythonCHB, line 14: you just call os.rename(), and it will raise an exception (IOError, maybe?) if the file isn't there -- remember IEAFP (It's Easier to Ask Forgiveness than Permission) -- plus, less code to write.

Students/SSchwafel/session06/functional_files.py
```diff
@@ -0,0 +1,59 @@
...
+
+        new_name = raw_input( "The current filename is {}, what would you like to call it instead? ".format(file_to_be_renamed))
+
+        os.rename(file_to_be_renamed, new_name)
+    
+    else:
+        print "It doesn't look like that file is present"
+
+map(renamer,filename)
```
PythonCHB, line 23: did this work?

you wouldn't usually use map for a function that required user input.

this looks like it would apply the renamer function to each letter in the filename.

PythonCHB:
html_render looks good  -- kind of like mine, so of course t does ;)

but if you did this mostly blind -- really nice job.

Students/SSchwafel/session06/functional_files.py
saschwafel, line 23: Yes, actually this did work. I am actually not sure the benefit of using map() in this instance, you can see my abandoned code at the bottom (which also worked).

PR #75, opened 2014-11-18
Session06 assignment

imdavis:
Solutions + tests to the html render assignment and the script to clean the leading and trailing whitespace of text in a file.


PythonCHB:
I like the incremental committing -- nice!

Students/imdavis/session06/cleanfile.py
```diff
@@ -0,0 +1,29 @@
...
+# cleans a file of leading and trailing whitespace
+
+import sys
+
+def strip_whitespace(astring):
+    """
+    Simple function that takes a string as an argument and cleans
+    leading and trailing whitespace
+    """
+    return astring.lstrip().rstrip()
```
PythonCHB, line 12: str.strip() does both in one call ;-)

Students/imdavis/session06/cleanfile.py
```diff
@@ -0,0 +1,29 @@
...
+readfile = open(sys.argv[1], 'r') # file name to be cleaned
+# new file name to write to
+newfile = open(raw_input("'(N)ew clean file name' ?> "), 'w')
+
+lines = []
+for line in readfile:
+    # add each new line in the file to the list
+    lines.append(line)
+    # use a map to clean each element of leading and trailing whitespace
+    cleanlines = map(strip_whitespace, lines)
```
PythonCHB, line 23: you got the map call right.

But since a file object is an iterator, you can actually do just:

cleanlines = map(strip_whitespace, readfile)

Students/imdavis/session06/test_html_render.py
```diff
@@ -0,0 +1,179 @@
...
+Some unit testing for our HTML renderer
+
+Make sure that pytest is installed.
+Since the name of this testing script file begins with 'test_'
+pytest will find it and run it.
+
+In the directory where this file lives, simply do:
+    $ py.test
+    -or-
+    $ py.test -s (will show print statements)
```
PythonCHB, line 13: nice find -- I had forgotten about that option.

PythonCHB:
looking good - nice work!

PR #77, opened 2014-11-18
Session 6 HW Add

dfugelso:


Students/Dave Fugelso/Session 6/html_render.py
```diff
@@ -0,0 +1,316 @@
...
+
+
+
+class Element(object):
+    '''
+    Step One, create the element class.
+    '''
+    tag = 'html'
+    indent = '   ' 
+    attributes = ''
```
PythonCHB, line 19: no reason to have an empty class attribute for this -- every instance willl get its own.

Students/Dave Fugelso/Session 6/html_render.py
```diff
@@ -0,0 +1,316 @@
...
+    '''
+    tag = 'body'
+    
+class pElement (Element):
+    '''
+    Sub class Element to replace tag 'html' with 'p' (paragraph)
+    '''
+    tag = 'p'
+    
+class headElement (Element):
```
PythonCHB, line 66: this one should be a bit more -- a header line can be H1, H2, etc....

Students/Dave Fugelso/Session 6/html_render.py
```diff
@@ -0,0 +1,316 @@
...
+    '''
+    Re-defines render top print on one line.
+    '''
+    def render (self, file_out, ind = ''):
+        '''
+        Render the line to file_out.write.
+        '''
+
+        for content in self.contents:
+            file_out.write (ind+'<'+self.tag+'>'+content+'</'+self.tag+'>'+'\n')
```
PythonCHB, line 82: are the attributes getting written out here? it doesn't look like it.

Students/Dave Fugelso/Session 6/html_render.py
```diff
@@ -0,0 +1,316 @@
...
+    
+#Step 5
+
+class SelfClosingTag(Element):
+    '''
+    Uses Element's init and replaces render with self closer.
+    '''
+    def render (self, file_out, ind = ''):
+        if len(self.contents) > 0:
+            file_out.write (ind+'<'+self.tag+' '+self.contents[0]+'/>\n')
```
PythonCHB, line 98: only the first element of contents?

Students/Dave Fugelso/Session 6/html_render.py
```diff
@@ -0,0 +1,316 @@
...
+    '''
+    tag = 'a'
+    def __init__(self, url, link):
+        self.contents = [' href="{}">{}'.format(url,link)]
+    
+    def render (self, file_out, ind = ''):
+        '''
+        Render the line to file_out.write.
+        '''
+        file_out.write (ind+'<'+self.tag+self.contents[0]+'/>\n')
```
PythonCHB, line 123: why contents in alist if you only want the one item anyway?

but nice -- I think  bit cleaner than how I did it.

though does HTML allow other arbitrary attributes in a Anchor tag? IF so you don't support that here.

Students/Dave Fugelso/Session 6/html_render.py
```diff
@@ -0,0 +1,316 @@
...
+    tag = 'ul'
+
+class liElement (Element):
+    tag = 'li'
+    
+class headerElement(OneLineTag):
+    '''
+    Create tag based on header level. Redefines __init__ and calls base class __init__.
+    '''
+    def __init__ (self, level, content=None):
```
PythonCHB, line 137: there we go! perfect.

Students/Dave Fugelso/Session 6/html_render.py
```diff
@@ -0,0 +1,316 @@
...
+    
+if __name__ == '__main__':
+    testStepOne()
+    testStepTwo()
+    testStepThree()
+    testStepFour()
+    testStepFive()
+    testStepSix()
+    testStepSeven()
+    testStepEight()
```
PythonCHB, line 316: nice work!

do try to make some unit tests at some point -- it can be very helpful.

Students/Dave Fugelso/Session 6/lambda_lab.py
```diff
@@ -0,0 +1,57 @@
...
+8
+### If you loop through them all, and call them, each one adds one more
+to the input, 5... i.e. the nth function in the list adds n to the input.
+
+'''
+
+def function_build(n):
+    l = []
+    for  i in range(n):
+        l.append(lambda x,j=i:x+j)
```
PythonCHB, line 38: you got it.

PythonCHB:
You generally wouldn't want to put all the test output files in git.

But other than that, really nice work.

PR #78, opened 2014-11-20
Homework 7

voite1:
Added Homework 7

### DIFF
--- a/Students/A.Kramer/session07/circle.py
+++ b/Students/A.Kramer/session07/circle.py
@@ -1,0 +1,101 @@
+'''
+Created on Nov 18, 2014
+
+@author: Aleksey
+'''
+
+from math import pi
+
+class Circle(object):
+    def __init__(self, the_radius):
+        self._radius = float(the_radius)
+        
+    @property
+    def radius(self):
+        return self._radius
+    
+    @radius.setter
+    def radius(self, value):
+        self._radius = value
+    
+    @property
+    def diameter(self):
+        return self._radius * 2.0
+    
+    @diameter.setter
+    def diameter(self, value):
+        self._radius = value / 2.0
+    
+    @property
+    def area(self):
+        return pi * self._radius**2.0
+    
+    @classmethod
+    def from_diameter(cls, diam):
+        return cls(diam / 2.0)
+
+    def __str__(self):
+        return "Circle with radius: %s" % self.radius
+    
+    def __repr__(self):
+        return "Circle(%s)" % self.radius
+    
+    def __add__(self, cls):
+        if isinstance(cls, Circle):
+            self.radius = self.radius + cls.radius
+        return self.radius
+    
+    def __mul__(self, num):
+        self.radius = self.radius * num
+        return self.radius
+    
+    def __eq__(self, other):
+        if isinstance(other, Circle):
+            return self.radius == other.radius
+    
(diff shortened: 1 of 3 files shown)


Students/A.Kramer/session07/circle.py
```diff
@@ -0,0 +1,101 @@
...
+'''
+
+from math import pi
+
+class Circle(object):
+    def __init__(self, the_radius):
+        self._radius = float(the_radius)
+        
+    @property
+    def radius(self):
```
PythonCHB, line 14: Here's the beauty of properties: if all you are doing is getting and setting the value -- no reason to make it a property, just have it be a regular old attribute.

If you need to refactor it later to do something, you can make it a property then, and the API won't change.

Students/A.Kramer/session07/circle.py
```diff
@@ -0,0 +1,101 @@
...
+    def diameter(self):
+        return self._radius * 2.0
+    
+    @diameter.setter
+    def diameter(self, value):
+        self._radius = value / 2.0
+    
+    @property
+    def area(self):
+        return pi * self._radius**2.0
```
PythonCHB, line 31: probably want: self._radius**2 -- I"m not entirely sure, but I suspect python goes through a faster routine for an integer power.

Students/A.Kramer/session07/circle.py
```diff
@@ -0,0 +1,101 @@
...
+        return cls(diam / 2.0)
+
+    def __str__(self):
+        return "Circle with radius: %s" % self.radius
+    
+    def __repr__(self):
+        return "Circle(%s)" % self.radius
+    
+    def __add__(self, cls):
+        if isinstance(cls, Circle):
```
PythonCHB, line 44: Do you want None if it's not an instance? 

better to raise a TypeError -- or just not bother checking.

Students/A.Kramer/session07/circle.py
```diff
@@ -0,0 +1,101 @@
...
+    def from_diameter(cls, diam):
+        return cls(diam / 2.0)
+
+    def __str__(self):
+        return "Circle with radius: %s" % self.radius
+    
+    def __repr__(self):
+        return "Circle(%s)" % self.radius
+    
+    def __add__(self, cls):
```
PythonCHB, line 43: I wouldn't use 'cls' here -- any name will do, but 'cls' is often used for class objects in class methods -- I think the convention is "other"

Students/A.Kramer/session07/circle.py
```diff
@@ -0,0 +1,101 @@
...
+    
+    def __lt__(self,other):
+        if isinstance(other, Circle):
+            return(self.radius < other.radius)
+
+    def __gt__(self,other):
+        if isinstance(other, Circle):
+            return(self.radius > other.radius)
+
+# did not do #9 - could not figure it out.
```
PythonCHB, line 64: you need to use the "Reflected arithmetic operators":

**radd**(self, other)

though I need to test and see how that works with something that also defines **add**...

Students/A.Kramer/session07/test_Circle.py
```diff
@@ -0,0 +1,86 @@
+'''
```
PythonCHB, line 1: Thanks for humoring me and writing format unit tests!

PR #79, opened 2014-11-22
Added reflection

voite1:


PythonCHB:
Nice! 

And I see you discovered abstract base classes, too.

Do be careful though, do you want to be able to add a circle to either another circle or a number?

PR #80, opened 2014-11-23
Session 07 Assignment: Circle Class

imdavis:
The circle module does what it should, as far as I can tell.  Passes all of my tests in the "test_circle.py" pytest module.  I used assertions in the radius.setter and diameter.setter to throw an Assertion if someone tries to initialize a Circle object with a negative or zero radius (or diameter), since that doesn't make sense.  

Not sure if this is right way to do this.  I remember Chris saying something about how assertions may not be the right way to handle this when he was discussing the "vector.py" example.  Something about how the assertions would be stripped out if the code was run in optimized mode.  Would appreciate feedback on if there is a more robust way such a thing should be handled.


Students/imdavis/session07/circle.py
```diff
@@ -0,0 +1,109 @@
...
+    def __init__(self, radius):
+        self.radius = float(radius)
+
+    @property
+    def radius(self):
+        return self.__radius
+
+    @radius.setter
+    def radius(self, radius):
+        assert radius > 0, "radius must be nonzero and non-negative"
```
PythonCHB, line 19: you are right -- assert is not the right way to do this. you want something like:

if radius > 0:
    raise ValueError("radius must be nonzero and non-negative")

but nice job structuring it for this kind of check.

Students/imdavis/session07/circle.py
```diff
@@ -0,0 +1,109 @@
...
+        Make sure we can multiply two Circle objects or a number and a 
+        Circle object.  In the case of multiplying a Circle object and a
+        number, this function will only be able to handle something 
+        like:
+            c = Circle(2.34)
+            c * 1.234
+        Need to define __rmul__ special method (below) for swapped 
+        operands.
+        """
+        if isinstance(c, Circle):
```
PythonCHB, line 60: nice!

Students/imdavis/session07/circle.py
```diff
@@ -0,0 +1,109 @@
...
+            return Circle(self.radius * c.radius)
+        else:
+            return Circle(c * self.radius)
+
+    def __rmul__(self, c):
```
PythonCHB, line 65: nice find.

PythonCHB:
really nice work!

-Chris

PR #82, opened 2014-11-24
Session7 Homework

Anastomose:
Circle assignment is completed and functional.  Additional updates to render may be submitted under Session6 branch.


.gitignore
```diff
@@ -1,6 +1,22 @@
 slides_sources/build
```
PythonCHB, line 1: this is a trick -- if I accept this, then I get your .gitignore -- but that's not a good one for the whole project -- it ignores some stuff I need to have in the repo.

I"ll see if I can figure out how to merge this right -- sone of this is stuff we will all want to ignore.

Students/ebuer/session_07/circle.py
```diff
@@ -0,0 +1,77 @@
+
+"""
+circle.py
+"""
+
+from math import pi
+import functools  # imported to access @total_ordering
+
+
+@functools.total_ordering
```
PythonCHB, line 10: nice find !

Examples/Session06/html_render/test_html_output1.html
```diff
@@ -1,4 +0,0 @@
-    <html>
```
PythonCHB, line 1: ahh -- and these too -- they should be left here.

rule of thumb -- don't mess with the Examples or Solutions dirs -- only put stuff in your own dir in Students.

Solutions/Session06/run_html_render.py
```diff
@@ -11,7 +11,7 @@
```
PythonCHB, line 11: also this -- you want o copy it to your students dir before making changes -- or don't "commit -a"

PythonCHB:
See if you can clean up this PR so that project files that aren't your aren't touched.

PR #83, opened 2014-11-24
Session 07 Assignment: Circle Class

imdavis:
Updated circle.py and test_circle.py to use a "raise ValueError" instead of an assertion to prevent an instantiation of a Circle object with a negative or zero radius (or diameter).  Thanks Chris!


codedragon:
Looks good!

PR #84, opened 2014-11-24
Added Session 07 Homework

salimhamed:
Here's my session 07 homework.  Please review and provide your feedback.
- Salim


Students/salim/session05/email/Salim Sr.txt
```diff
@@ -0,0 +1,9 @@
+
```
PythonCHB, line 1: Note: you generally don't want to add generated files to git....

Students/salim/session08/sparse_array.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+class SparseArray(object):
+    """
+    Use a dictionary to store the location of the values.  Then, if the user
+    asks for a key that isn't in the dictionary, but it should exist, then
+    return zero.
+    """
+    def __init__(self, sequence):
+        self.data = {k: v for k, v in enumerate(sequence) if v != 0}
```
PythonCHB, line 11: nice use of dict comprehension

Students/salim/session08/sparse_array.py
```diff
@@ -0,0 +1,76 @@
...
+                        d[k] = v
+                self.data = d
+        else:
+            raise IndexError
+
+    def __iter__(self):
+        self.current = -1
+        return self
+
+    def next(self):
```
PythonCHB, line 71: nice!

Students/salim/session09/decorator_lab.py
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env python
+
+
+def p_wrapper(func):
+    def wrap(*args, **kwargs):
```
PythonCHB, line 5: Not quite what I was looking for. In this case, you would need to  pass the tag into the decorated function. I was thinking that yo would pass the tag into the decorator itself -- the function being decorated doesn't need to know anythign about tags.

@add_tag(tag='div')
def some_func_that_returns_a_string():
    ....

PR #85, opened 2014-11-26
HMWK from 6 & 7 also initial code for class project

ARSimmons:
In HMWK 6 I am still having trouble getting the 'link' for 'google' to render correctly...


Students/ASimmons/Session06/html_render.py
```diff
@@ -0,0 +1,128 @@
...
+
+
+    def render(self, file_out, current_ind=""):
+        """
+        render, just renders itself
+        """
+        # from kwargs...need to format a dictionary of items, replaced with render_tag
+        #attrs = "".join([' {}="{}"'.format(k, v) for k, v in self.kwargs.items()])
+
+        # file_out -- File name in which to output render
```
PythonCHB, line 50: just to be pedantic -- file_out is not a file name -- it is a file-like object (i.e. open file or other).

Students/ASimmons/Session06/html_render.py
```diff
@@ -0,0 +1,128 @@
...
+class Hr(SelfClosingTag):
+    tag = 'hr'
+
+
+class A(OneLineTag):
+    tag ='a'
+    # it is content, and not self.content because you
+    # haven't assigned it
+    def __init__(self, link, content, **kwargs):
+        OneLineTag.__init__(self, content=None, **kwargs)
```
PythonCHB, line 101: I think you need to pass content on through to OneLineTag.**init**-- rather than None.

PR #86, opened 2014-11-27
Homework 8 complete

voite1:
Completed HW 8.  Will start working on final project later this week.

### DIFF
--- a/Students/A.Kramer/session08/generator_solution.py
+++ b/Students/A.Kramer/session08/generator_solution.py
@@ -1,0 +1,106 @@
+'''
+Created on Nov 26, 2014
+
+@author: Alekesey Kramer
+'''
+
+# Generator to add numbers to predecessors
+def intsum():
+    num = 0
+    count = 0
+    while num >= 0:
+        yield num
+        count += 1
+        num += count
+
+
+def doubler():
+    num = 1
+    while num >= 0:
+        yield num
+        num = num*2
+        
+def prime():
+    # My code :-)
+    count = 2
+    while True:
+        if __is_prime(count):
+            yield count
+            count += 1
+        else:
+            count += 1
+
+# Found this code on the web at 
+# https://www.daniweb.com/software-development/python/threads/70650/an-isprimen-function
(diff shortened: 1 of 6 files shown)


Students/A.Kramer/session08/generator_solution.py
```diff
@@ -0,0 +1,106 @@
...
+            yield count
+            count += 1
+        else:
+            count += 1
+
+# Found this code on the web at 
+# https://www.daniweb.com/software-development/python/threads/70650/an-isprimen-function
+# with combination from here: http://stackoverflow.com/questions/15285534/isprime-function-for-python-language
+# and modified a bit to reject all the numbers lees than 2 (just in case I will use it later)
+# quite useful and fast function, but limited by the have the size of int...
```
PythonCHB, line 37: actually, python "int" is a "long int" -- i.e. it automatically rolls over to a unlimited precision integer -- so only limit is computation time.

But there is a bug -- it reports True for some even numbers::
  In [8]: for i in range(10):
     ...:     print i, __is_prime(i)
     ...:  
  0 False
  1 False
  2 True
  3 True
  4 True
  5 True
  6 True
  7 True
  8 True
  9 False

Students/A.Kramer/session08/generator_solution.py
```diff
@@ -0,0 +1,106 @@
...
+    num = 1
+    while num >= 0:
+        yield num
+        num = num*2
+        
+def prime():
+    # My code :-)
+    count = 2
+    while True:
+        if __is_prime(count):
```
PythonCHB, line 27: ideally, with a generator, you generate as you go -- so you don't want to do the whole is_prime calculation for each number. That is there is some information you already know when generating the next prime number, having computed a bunch so far. But you've got the key structure for generators down, which is the point here.

Students/A.Kramer/session08/generator_solution.py
```diff
@@ -0,0 +1,106 @@
...
+            yield count
+            count += 1
```
PythonCHB, line 29: if you are going to increment count either way, why not put it outside the if block?

PR #87, opened 2014-12-02
Added Generator Lab and Start on SparseArray

saschwafel:
Wasn't able to make much progress on the homework this week, I got the majority of the generator lab working, but am pretty confused about the rest of the assignments. I believe we were going over the sparse array lab in class, the notes of which are in my version, but I am still a bit lost. 

Also added some minor progress in my project, accessing and printing from the API

### DIFF
--- a/Students/SSchwafel/session08/generator_lab.py
+++ b/Students/SSchwafel/session08/generator_lab.py
@@ -1,0 +1,74 @@
+#!/usr/bin/python
+
+
+def sum_generator(number_range):
+
+    counter = list(range(number_range))
+
+    new_value = 0
+    i = new_value
+
+    while i < number_range:
+
+	try:
+	
+	    item = counter.pop(0)+counter.pop(0)
+	    yield item
+	    counter.insert(0,item)
+
+	except IndexError:
+
+	    break
+
+
+sum_gen = sum_generator(10)
+print next(sum_gen)
+print next(sum_gen)
+print next(sum_gen)
+print next(sum_gen)
+print next(sum_gen)
+print next(sum_gen)
+
+def doubler(number_range):
+    i = 0
+    while i<number_range:
+	if i < 1:
+	    i = i+1
+	    yield i
+	    
+	else:
+
+	    yield i*2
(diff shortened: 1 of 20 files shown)


Students/SSchwafel/session08/generator_lab.py
```diff
@@ -0,0 +1,74 @@
+#!/usr/bin/python
+
+
+def sum_generator(number_range):
+
+    counter = list(range(number_range))
```
PythonCHB, line 6: range() returns a list -- no need to make a a list out of it.

(maybe you've seen python3 code, where range() returns an iterator...)

Students/SSchwafel/session09/p_wrapper_lab.py
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/python
+
+def p_wrapper_func(func, tag):
+    
+    def tagger(tag):
```
PythonCHB, line 5: what's this for???

But the rest of it looks good.

Students/SSchwafel/session10/unicode_lab.py
```diff
@@ -0,0 +1,21 @@
...
+
+unicode_chess_piece = u'\u2654'
+
+unicode_chess_piece = unicode_chess_piece.encode('utf-8')
+
+print unicode_chess_piece
+
+#print unicode_string
+
+print file.read()
```
PythonCHB, line 15: This is going to give you a byte string with the encoded data in it --probably not what you want.

try:

import io

file = io,open(filename, encoding='utf-8')

then you can read and write to it and it will handle the encoding/decoding for you.

Students/SSchwafel/session10/unicode_lab.py~
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/python
```
PythonCHB, line 1: this is a back-up file -- try not to commit those to git.

git add \* is dangerous!

PythonCHB:
I'll commit this, and review the project over the weekend.

PR #88, opened 2014-12-06
Session 09

Anastomose:
Session 9 merge request.  Managed to get decorator class with arguments working after some googling.  Still sort of abstract though.


.gitignore
```diff
@@ -1,6 +1,25 @@
 slides_sources/build
```
PythonCHB, line 1: this is going to override my .gitignore.

In theory, adding .gitignore to the .gitignore should allow you to not override mine. I think what you may need to do is delete .gitignore, then checkout mine again, add .gitignore to that, and see what happens!

Students/ebuer/session_09/p_wrapper.py
```diff
@@ -0,0 +1,38 @@
...
+# dummy func to return a string
+
+def p_wrapper(func):
+    def wrapped(*args):
+        result = func(*args)
+        return '<p> {} </p>'.format(result)
+    return wrapped
+
+
+class tag_wrapper(object):
```
PythonCHB, line 15: nicely done!

PythonCHB:
I'm going to merge and try to clean up the .gitignore on my end.

PR #90, opened 2014-12-07
Sessions 7, 8, 9 Homework

dfugelso:
Homework for the last three sessions (7,8,9)


Students/Dave Fugelso/Session 7/Circle.py
```diff
@@ -0,0 +1,65 @@
...
+    def __str__(self):
+        return 'Circle wih radius {}'.format(self.radius)
+ 
+    def __add__(self, other):
+        if isinstance(other, Circle):
+            return Circle (self.radius + other.radius)
+        else:
+            self.radius += other
+            return self
+
```
codedragon, line 46: You need to be consistent here. In one case you are returning a new object, and in the other you are returning the original object changed. You should almost certainly be returning a new object. If you want a function that returns the same object changed, you would use the augmented assignment (**iadd**, for example). You are correct that the reflected operator (**rmul**, below) would return the same thing as the normal operator (here and in most cases), and is often just a call to its non-reflected equivalent.

Students/Dave Fugelso/Session 7/circle_test.py
```diff
@@ -0,0 +1,81 @@
+'''
+These are all the unit test to make sure circle works as expected.
+
+Really taking these seriously as feedback from last homework assignment indicated I should.
+I skipped the unit tests and did step tests. Not this time!
+'''
+
```
codedragon, line 7: Yay, testing! You will really find this helpful when you are refactoring code and adding features. Makes it much easier to figure out if you broke old functionality making changes, and will give you hints as to why it happened.

Students/Dave Fugelso/Session 8/Sparse.py
```diff
@@ -0,0 +1,80 @@
...
+
+Decisions:
+
+1. Init with just a length.
+2. May explicitly change size with setSize. Deletes entries out of range on shrink.
+3. Use set to populate array
+
+Implement with dictionary
+
+'''
```
codedragon, line 20: Looks like you mostly figured it out. You need to be careful, however, about what it means to delete a number in your array. In your implementation, you are changing that number to a zero. This may be what you want, but you should be explicit in your documentation that is what is happening. Many people would expect that number to go away completely, and the array to become one shorter in length, meaning that any array entries greater than the one deleted will change positions by one.

Students/Dave Fugelso/Session 8/generator_solution.py
```diff
@@ -0,0 +1,104 @@
...
+    while True:
+        sum = i + sum
+        yield sum
+        i += 1
+
+
+def intsum2():
+    '''
+    test_generator has a intsum2... dunno why
+    '''
```
codedragon, line 31: There are at least a couple of ways to implement this, and Chris was just looking to see if students could come up with another way. The most efficient way in your case is:
def intsum2():
    return intsum()
Amusing myself there, hope you don't mind.

Students/Dave Fugelso/Session 8/iterator.py
```diff
@@ -0,0 +1,8 @@
+'''
+Extend iterator_1 to do xrange-like iteration
+'''
+
+from iterator_1 import IterateMe_1 
+
+class iterator (IterateMe_1):
+
```
codedragon, line 8: Doesn't seem to do much...

Students/Dave Fugelso/Session 8/test_iterator.py
```diff
@@ -0,0 +1,14 @@
...
+
+from iterator import Sparse
+import unittest
+
+def test_iteration(self):
+    for i in iterator_inst(0, 50, step=1):
+        
+    
+if __name__ =='__main__':
+    unittest.main()
```
codedragon, line 14: Nice test. ;-) I am hoping you just ran out of time, and not that you are/were confused.

codedragon:
Made some comments on homework. Will send an email about project, since we have an updated version.

PR #92, opened 2014-12-10
WFukuhara Final Project

WayneUW:
Hi Chris - Pull request for final project. I also removed a lambda_keyword.pyc from the slides_sources / week-06 folder because I got a git message it had been modified. Not sure what was going on with that.


Students/WFukuhara/Final/Fukuhara_final.py
```diff
@@ -0,0 +1,85 @@
...
+        3 - The avg monthly low for the last 20 years
+        4 - (Select this option when you are finished)
+        ''' )
+    return uentry
+
+
+def highlowresults(a, b):   # function with header and row as arguments
+    """ takes the High or Low column data for each month for the past 20 years and calcuates the average price.
+    """
+    fin = open( 'disneystock.csv', 'U' )
```
codedragon, line 30: It would be lovely to have the filename be a changeable input. Flexibility is good.

Students/WFukuhara/Final/Fukuhara_final.py
```diff
@@ -0,0 +1,85 @@
...
+    avghigh = highlowresults( 'High', 2 )    # assign the return results
+    print avghigh    # this is a more flexible alternative to having the print statement at the end of the highlowresults function. 
+
+
+def lowavg( ):
+    """ contains the 'Low' column index and format parameters
+    """
+    avglow = highlowresults( 'Low', 3 )
+    print avglow
+
```
codedragon, line 67: It looks like for each of your menu items, you are re-reading the file. This can be pretty costly, especially with large files. It would be more efficient to save the data you care about, and just run the analysis for the particular menu item. You could get even fancier, and save the results when a particular analysis is run, so you don't even have to re-run the analysis for subsequent requests.

Students/WFukuhara/Final/Fukuhara_final.py
```diff
@@ -0,0 +1,85 @@
...
+    fin1 = open('disneystock.csv', 'U')
+    finrow1 = csv.reader(fin1)    # converts each row to  list
+    print "Here are the monthly open and close prices for the last 20 years:"
+    finrow1.next( )    # skip over the iniitial header row
+    for row1 in finrow1:
+        out.write( '%10s %s %.2f %s %.2f' % ((row1[0]), '     ', float(row1[1]), '     ', float(row1[4])) + '\n' )   #  format with 2 decimal places.
+
+    fin1.close( )
+
+
```
codedragon, line 54: Highavg and lowavg are basically the same method, and don't seem to add anything new. Why not just call highlowresults directly from the menu with different inputs?

Students/WFukuhara/Final/test_Fukuhara_final.py
```diff
@@ -0,0 +1,56 @@
...
+test harness for Fukuhara_final.py
+
+"""
+
+import sys
+from StringIO import StringIO
+from Fukuhara_final import highlowresults
+from Fukuhara_final import opencloseresults
+
+
```
codedragon, line 11: Yay, tests! I find it helpful to have long, explicit names for my tests, since you pretty much never run them directly. For example, your first test could be test_highlow_returns_correct_string. You have actually written potential names in your comments, making these the test names means you don't actually even have to look in the test file when a test fails, you just know that test_correct_result_for_avg_high (as opposed to test2_avgresults) didn't work, and figure out why.

codedragon:
Nice work! Looks like you have learned some Python. See comments in the diffs. Hope to see you next quarter. cheers, Maria

PR #94, opened 2014-12-11
Final project

gosunfish:
I will be continuing to work on the final project, but wanted to get something submitted now in case the power goes out from the wind storm that has been predicted.


Students/CarolynEvans/final_project/config.py
```diff
@@ -0,0 +1,42 @@
...
+        :return: There is no return value. This method is called during __init__.
+        """
+        # TODO: Rewrite for loop as comprehension
+
+        for key, val in config.items():
+            if type(val) is dict:
+                self.flatten(val)
+            else:
+                self._VALUES[key]=val
+
```
codedragon, line 31: Can you be sure that a dictionary inside a dictionary does not contain a key of the same name as an outside dictionary? Should you be checking for this?

Students/CarolynEvans/final_project/data_warehouse.py
```diff
@@ -0,0 +1,54 @@
...
+                    return records
+                except ProgrammingError as e:
+                    # This error occurs when there are no records to return,
+                    # for example, queries such as inserts, updates, and unloads.
+                    message = unicode(e)
+                    if message == u'no results to fetch':
+                        return message
+                    else:
+                        raise message
+
```
codedragon, line 52: Nice use of try and except.

Students/CarolynEvans/session05/mailroom.py
```diff
@@ -0,0 +1,26 @@
...
+
+        if menu_choice == 1:
+            send_thanks()
+
+        elif menu_choice == 2:
+            create_report()
+
+if __name__ == '__main__':
+    main()
+
```
codedragon, line 26: Nice start. We were hoping for a little more functionality out of our mailroom, though.

codedragon:
Great start Carolyn. You have demonstrated you can use some Python, and I look forward to seeing more from you next quarter.

PR #95, opened 2014-12-12
Term Project Pull Request

Anastomose:
Term Project for IntrotoPython Class 1; custom class for reading and plotting validated environmental data.


Anastomose:
Did a git rm .gitignore followed by pull then reset HEAD.  Looks like pull request can be automatically merged now.

Students/ebuer/termp/test_termp_funcs.py
```diff
@@ -0,0 +1,88 @@
...
+    p_test = callclass()
+
+    junk_dict = {'name': 'samples', 'o_list': p_test.samples.unique()}
+    p_test.showOptions(**junk_dict)
+
+    two_dict = {'name': 'methods', 
+                'o_list': ['Method A', 'Method B', 'Method C']}
+    p_test.showOptions(**two_dict)
+    assert True
+
```
codedragon, line 44: Most straight-forward way I've found to test print is using standard out:

def foo(out=sys.stdout):
    out.write("hello, world!")

Then the test is much simpler:

def test_foo():
    from foomodule import foo
    from StringIO import StringIO

```
out = StringIO()
foo(out=out)
output = out.getvalue().strip()
assert output == 'hello world!'
```

Students/ebuer/termp/test_termp_funcs.py
```diff
@@ -0,0 +1,88 @@
...
+    p_test = callclass()
+    p_test.selectSample('14051305')
+    p_test.selectForPlot('SW8082A')
+    # pdb.set_trace()
+    sample_dict = p_test.makePlotobj()
+
+    f = sample_dict['flag']
+    for fl in f:
+        print type(fl)
+    assert True
```
codedragon, line 88: Looks like you just needed to know what types were in your flag value, but didn't need to test anything further? Might not want to keep this in your test file, unless you want to verify the types, in which case you should do that. Nice to keep test files pretty clean, otherwise if other people use them they get confused.

Students/ebuer/termp/createplot.py
```diff
@@ -0,0 +1,98 @@
...
+   This file can be further refined in the future to initialized with a 
+   datafile create a full dictionary of samples and then plot each
+   sample/method pair in sequence. 
+   """
+
+import sys
+sys.path.append(r"D:\GitHub\IntroToPython\Students\ebuer\termp")
+
+import termp_funcs
+reload(termp_funcs)
```
codedragon, line 14: Why do you reload? Usually this is used to re-import a module in the python interpreter after a file has changed.

Students/ebuer/termp/createplot.py
```diff
@@ -0,0 +1,98 @@
...
+# create a function to make use of the plot_dict
+
+
+def plotSampdict(**kwargs):
+    """Takes plotting dictionary from plotprep and returns figure
+
+       Not sure how to test this function since most of what is
+       rendered goes to the screen.
+       """
+
```
codedragon, line 55: Not usually necessary to test outside modules, like matplotlib, so just make sure the values you are sending are what you expect, and the graph looks like you expect in the end, and should be good.

Students/ebuer/termp/termp_funcs.py
```diff
@@ -0,0 +1,138 @@
...
+        # filter self.sampleData by method and return to class attributes
+        mask = self.sampleData['std_anl_method_name'].isin([methodval])
+        self.plotData = self.sampleData[mask]
+
+        return self.plotData
+
+
+    def makePlotobj(self):
+        """Use method-subsetted data and make a plotting object (dict?)"""
+
```
codedragon, line 113: Yes, you are returning a dictionary, and calling it an object is confusing.

Students/ebuer/termp/termp_funcs.py
```diff
@@ -0,0 +1,138 @@
...
+"""
+
+import pandas as pd
+import pdb
+
+
+class plotprep(object):
+    """Class is initialized with validated data file and used to create plotting object"""
+
```
codedragon, line 18: By convention, classes are generally capitalized in Python.

Students/ebuer/termp/createplot.py
```diff
@@ -0,0 +1,98 @@
...
+import termp_funcs
+reload(termp_funcs)
+
+from termp_funcs import plotprep
+
+import pdb
+import pandas as pd
+import matplotlib.pyplot as plt
+
+
```
codedragon, line 22: Separate your functions and your script. If you leave the functions on the top, and collect the script in an 

if **name** == "**main**": 

block, it not only improves readability, but also allows you to import the functions elsewhere, without dealing with the script part. Iow, if you want to use these functions later, without running the script, you can just import the module, but you can still run the module as a program, as you do here.

codedragon:
This looks great, nice work. You have demonstrated you can write some Python. Hope to see you next quarter!

Students/ebuer/termp/createplot.py
Anastomose, line 14: I'm using Anaconda's ipython, once I import something into a session the module remains unchanged unless I explicitly reload it. Sort of an irritating feature really, but this seemed to solve the problem.

Students/ebuer/termp/test_termp_funcs.py
Anastomose, line 44: Thanks, this is very good, I will have to use it in the future.

Students/ebuer/termp/termp_funcs.py
Anastomose, line 113: I should have removed that note, it something to remind myself what format I wanted during development.

Anastomose:
I'll be there, see you then.

PythonCHB:
too late for that class -- sorry it wont' be in the tagged version in the repo.

PR #96, opened 2014-12-13
final project @ brydavis/final/json2sql.py

ghost:


PythonCHB:
Bryan,

Something went weird here -- a huge number of files have changes -- many of them just changes to the file mode. but i can't merge this as is at all. I"ll take a look at your code in your rep, though.

PythonCHB:
Bryan,

Neat idea, nice job.

A few comments:
- You've got tab indentation -- believe me, you really don't want to do that with Python.
- you've got a bunch of functions defined inside the reduce_dict function. You can do that, but you usually don't want to, and in this case, I don't see any reason to. It does mean that the definitions get re-run every time you call reduce_dict
- ```
    global model_dict
  ```
  
  You generally don't want o use globals like this. ANd in this case, it's totally unnecessary -- module level names are global anyway -- you can't re-bind them, but you can use them as you are here.
- put in some docstrings, so you can know what your code is doing.
- slick use of regex's -- now if only I could read them ;-)

not much more to say. Solid code.

Enjoy next class
 -Chris

openwonk:
Thanks for the feedback, Chris!  I enjoyed the class and look forward to seeing you in the spring.

Bryan

PR #97, opened 2014-12-13
Final file update for final class project. Sample data included.

ARSimmons:
Well...it runs. Looks like everything is going okay. So far. :) 

To be honest, I have some concerns that some of my unittests are passing (but shouldn't be). Earlier this week I had a tuple error in 'def cut_line_by_junctions'. i.e.:

for point in pointsLineList:
             # quantitize the point
             self.quantitize(point) 

...This error should've caused tests like 

def test_splitting_lines_2_arcs

&

def test_splitting_line_w_junction_at_start_1_arc

to fail...but, uh...it never did. I have been investigating this issue ever since. 

If you see a problem there feel free to call me on it. It has been driving me nuts!

Great class! Thanks!


PythonCHB:
merging to review -- it's easier that way!

Issue #115, opened 2015-09-22
Update How to Install Getting Python on Linux notes


PythonCHB:
slides_sources/source/supplements/python_for_linux.rst

PythonCHB:
Updated -- but I don't have a Linux box handy.

Could someone give this a look-see? 

Also, nice to support Ubuntu, and maybe an rpm-based system too?

ranchodeluxe:
@PythonCHB I've put in a pull request for these changes, just in case you didn't get notified about that

PythonCHB:
looks great -- not closing the issue until we make a decision about ensurepip.

ranchodeluxe:
@PythonCHB : let me look at `ensure-pip` and get back to you, never heard of it. It sounds a little to much like that drink

PythonCHB:
added ensure-pip (without testing) -- I guess we'll find out in class!

PR #127, opened 2015-09-30
#115: Update Linux Debian and CentOS Python Install Notes

ranchodeluxe:
@ChrisBarker-NOAA 
I've tested these out on Ubuntu15.04 and CentOS7. Let me know what you think.

### DIFF
--- a/slides_sources/source/supplements/python_for_linux.rst
+++ b/slides_sources/source/supplements/python_for_linux.rst
@@ -120,9 +120,9 @@ You should now be able to run ``iPython``::
 Fedora and Red Hat Related Distros (CentOS)
 ==================================================
 
-.. warning::
+.. raw::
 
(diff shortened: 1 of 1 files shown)


slides_sources/source/supplements/python_for_linux.rst
```diff
@@ -55,31 +91,120 @@ To get pip, the first option is to use your system package manager, something li
 
 .. code-block:: bash
 
-    $apt-get install python-pip
+    $ sudo apt-get install python3-pip
```
PythonCHB, line 94: Is ensure-pip not there is the the version installed by the package manager?
it looks like it's there in py3.4.3:

https://docs.python.org/3.4/library/ensurepip.html

if so, I'm still not sure if it's a better way to go that apt-get....

PythonCHB:
This looks great -- thanks!

Still wondering about ensurepip, but merging anyway.

PR #162, opened 2018-08-10
fix simple typo

lpmi-13:


PythonCHB:
Thanks!

Note that continued development of the program is happening in the PythonCertDevel repo:

https://github.com/UWPCE-PythonCert/PythonCertDevel

This repo may not see any future use. (though there is still some good stuff in there that isn't being used in the new program...)

PR #163, opened 2021-01-02
Revised the script

jerrytozhang:


PythonCHB:
This is from a class from 6 years ago -- I'm wondering why you are looking at it now ?!?

